PR #70, opened 2023-04-25
Update conformance-test report for https://github.com/jspecify/jspecify/pull/376

### LINKED ISSUE #17: Add support for @NonNull
The change should be a trivial addition to this list:
https://github.com/jspecify/nullness-checker-for-checker-framework/blob/02a51b070ece23700a617d3e5b89d0c42e1df514/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java#L1956

But it may be a good opportunity for someone to start playing around with building the checker, running its tests, and expanding the samples directory (currently used by its tests) to exercise `@NonNull`, particularly any interesting cases it makes possible (or at least makes more likely).

cpovirk:
Fixes #17

Replaces #63

### DIFF
--- a/tests/ConformanceTest-report.txt
+++ b/tests/ConformanceTest-report.txt
@@ -1,4 +1,4 @@
-# 28 pass; 547 fail; 575 total; 4.9% score
+# 28 pass; 545 fail; 573 total; 4.9% score
 FAIL: AnnotatedInnerOfNonParameterized.java: no unexpected facts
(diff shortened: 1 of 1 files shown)


cpovirk:
Yay, https://github.com/jspecify/jspecify-reference-checker/pull/79 matches—or rather, it's slightly different because it reflects the changes in line numbers from https://github.com/jspecify/jspecify/pull/387, which I would otherwise need to have regenerated this PR for.

PR #72, opened 2023-05-03
Adapt to spotless reformatting

wmdietl:
In PR https://github.com/jspecify/jspecify/pull/381


cpovirk (approved):
Thanks. I imagine that ConformanceTestOnSamples-report.txt will also need adjusting, but I haven't looked.

(will look at the actually reformatting soon—I guess I'll merge that first?)

PR #73, opened 2023-05-03
Adapt conformance results for reformatted samples.

cpovirk:
This is like #72 but for the samples, rather than the new
conformance-test directory.

### DIFF
--- a/tests/ConformanceTestOnSamples-report.txt
+++ b/tests/ConformanceTestOnSamples-report.txt
@@ -6,216 +6,216 @@ FAIL: AnnotatedTypeParameter.java: no unexpected facts
 FAIL: AnnotatedTypeParameterUnspec.java: no unexpected facts
 FAIL: AnnotatedWildcard.java: no unexpected facts
 FAIL: AnnotatedWildcardUnspec.java: no unexpected facts
-FAIL: ArraySameType.java:36 jspecify_nullness_mismatch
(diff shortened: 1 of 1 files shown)


wmdietl (approved):
Quick eye-balling looks good given the one-line change to many files.
I assume you created this using `JSPECIFY_CONFORMANCE_TEST_MODE=write ./gradlew conformanceTestOnSamples`. I had to look around again to find how to do this (I nearly submitted a PR for this myself just now, but noticed your PR just in time).
Should we document this in the top-level README?

cpovirk:
> I assume you created this using `JSPECIFY_CONFORMANCE_TEST_MODE=write ./gradlew conformanceTestOnSamples`. 

Yep, exactly.

> I had to look around again to find how to do this (I nearly submitted a PR for this myself just now, but noticed your PR just in time). Should we document this in the top-level README?

I assume that @netdpb has this among his plans (along with the automation that would save many people from having to worry about it at all). But I doubt he would object if you sent him a PR now :) (There may have been discussion of modifying the script to not use an environment variable? I forget. Worst case, we update the docs when we update the script.)

netdpb:
Haha. I'm testing my GitHub workflows to do just this, and I was wondering why everything was failing oddly.

I'm learning that the only way to test GitHub workflows is on the `main` branch of a private fork, which is what led me to push commits onto jspecify/main by mistake the other day.

PR #75, opened 2023-05-08
Workflow to create a PR to update conformance test reports

netdpb:
Workflow can be triggered manually or by webhook. The idea is for a workflow on https://github.com/jspecify/jspecify to trigger this one when a merge to `main` causes the conformance test reports to change.


cpovirk (approved):
Neat.

PR #80, opened 2023-05-22
Analyze each file in the top-level test directory by itself, instead of with its siblings.

netdpb:


src/test/java/tests/conformance/AbstractConformanceTest.java
```diff
@@ -190,9 +193,13 @@ private ImmutableList<ExpectedFactAssertion> readExpectedFacts(Path file) {
 
   private static final Pattern EXPECTATION_COMMENT = Pattern.compile("\\s*// (?<expectation>.*)");
 
-  private static ImmutableList<Path> javaFilesInDirectory(Path directory) {
+  private static ImmutableList<ImmutableList<Path>> javaFilesInDirectory(
```
cpovirk: It might be slightly simpler to return a `Stream` here, though I see that you'd still have to `collect` in the `!asIndividualFiles` case (or do something more clever than I can think of offhand.).

I'm also wondering if there's a clearer name for this method now that it returns a list of lists. "testFileGroupings" or something?

cpovirk (approved):

src/test/java/tests/conformance/AbstractConformanceTest.java
netdpb: I tried returning a `Stream`, but I don't think you can do that from within a try-with-resources, which `Files.list(Path)` must be in.

How about this? I've added a separate method to group the files using `Lists.partition` (and to filter out empty lists)

cpovirk: Oh, I see, thanks. LGTM.

cpovirk (approved):

PR #90, opened 2023-07-14
Update robot name and email to jspecify-bot.

netdpb:


.github/workflows/update-conformance-test-reports.yml
```diff
@@ -37,8 +37,8 @@ jobs:
           path: reference-checker
       - name: Create or update the Reference Checker ${{ env.branch }} branch
         run: |
-          git config user.name 'update-conformance-test-reports[bot]'
-          git config user.email 'update-conformance-test-reports[bot]@users.noreply.github.com'
+          git config user.name 'jspecify-bot'
```
cpovirk, line 40: [Dependabot uses "dependabot[bot]."](https://github.com/google/truth/commits?author=dependabot%5Bbot%5D) I wonder if that suffix is some kind of preferred pattern?

cpovirk (approved):

.github/workflows/update-conformance-test-reports.yml
netdpb: I don't know, but I think that has something to do with GitHub Apps?

Let's not do that unless we have to.

PR #96, opened 2023-08-18
Simplify test report

netdpb:
Improve "details" mode output so that unexpected facts are reported in-line with expected facts (with "OOPS" if the reported fact must be expected, or "INFO" if not). This way it will be easier to see what is going wrong when the comparison fails, and to see what other facts could be expected.

Make "details" mode also run the comparison and fail, just with extra output.

Also do some code simplification:
* `ConformanceTestReport` now holds all expected and reported facts and their mapping, rather than a set of passing and failing assertions.
* Instead of a `ConformanceTestSubject` that does complex comparison between the generated report and the stored one, use `Truth`'s very good text comparison to compare the text of the generated report to the text of the stored one.


src/test/java/tests/conformance/AbstractConformanceTest.java
```diff
@@ -263,225 +222,169 @@ static Mode fromEnvironment() {
...
 
     /** The file path relative to the test source root. */
     public final Path getFile() {
       return file;
     }
 
-    @Override
-    public final int compareTo(ConformanceTestAssertion that) {
-      return COMPARATOR.compare(this, that);
+    /** Returns the line number of the code in the source file to which this assertion applies. */
```
cpovirk:
```suggestion
    /** Returns the line number of the code in the source file to which this fact applies. */
```

src/test/java/tests/conformance/AbstractConformanceTest.java
```diff
@@ -263,225 +222,169 @@ static Mode fromEnvironment() {
...
+      return commentText;
+    }
+  }
 
-      /** Returns the fact expected at the file and line number. */
-      public ExpectedFact getFact() {
-        return fact;
-      }
+  /**
+   * An assertion that the tool behaves in a way consistent with a specific fact about a line in the
```
cpovirk, line 338: The relationship between `Fact`, `ExpectedFact`, and `ExpectedFactAssertion` isn't clear at first glance. That's partially about class naming (`ExpectedFact` is not a fact, but `ExpectedFactAssertion` is) and partially about how `ExpectedFact` and `ExpectedFactAssertion` have very similar Javadoc.

src/test/java/tests/conformance/AbstractConformanceTest.java
```diff
@@ -104,15 +97,20 @@ private ConformanceTestReport runTests() throws IOException {
...
   }
 
-  private Stream<ConformanceTestResult> analyzeFiles(List<Path> files) {
-    ImmutableMap<Path, ListMultimap<Long, ReportedFact>> reportedFactsByFile =
+  private Stream<ConformanceTestReport> analyzeFiles(List<Path> files) {
+    ImmutableListMultimap<Path, ReportedFact> reportedFactsByFile =
         Streams.stream(analyze(ImmutableList.copyOf(files)))
-            .collect(ReportedFact.BY_FILE_AND_LINE_NUMBER);
-    return files.stream().flatMap(file -> testResultsForFile(file, reportedFactsByFile).stream());
+            .collect(toImmutableListMultimap(ReportedFact::getFile, identity()));
```
cpovirk:
```suggestion
            .collect(toImmutableListMultimap(ReportedFact::getFile, rf -> rf));
```

mostly just for consistency with another file in the PR :)

src/test/java/tests/conformance/AbstractConformanceTest.java
netdpb: Yeah. I want to replace the current `ExpectedFact` with a string, and then rename `ExpectedFactAssertion` to `ExpectedFact` in a followup PR. Should I do that in this PR instead?

cpovirk: Followup is fine, thanks.

src/test/java/tests/conformance/AbstractConformanceTest.java
netdpb: I got rid of all the `identity()`s, and used `Multimaps.index()` to simplify further.

cpovirk (approved):
I still like `Multimaps.index`, too, and I may someday renew my advocacy for `collect(indexingBy(ReportedFact::getFile))` :)

PR #97, opened 2023-08-18
Check "sink-type" assertions

netdpb:
Partially satisfies #98.

### DIFF
--- a/src/test/java/tests/ConformanceTest.java
+++ b/src/test/java/tests/ConformanceTest.java
@@ -132,14 +132,18 @@ public final class ConformanceTest extends AbstractConformanceTest {
       }
       switch (detailMessage.messageKey) {
         case "sourceType":
-          return expressionType(
-              fixType(detailMessage.messageArguments.get(0)),
-              detailMessage.messageArguments.get(1));
-
+          {
+            String expressionType = fixType(detailMessage.messageArguments.get(0));
+            String expression = detailMessage.messageArguments.get(1);
+            return expressionType(expressionType, expression);
+          }
         case "sinkType":
-          return sinkType(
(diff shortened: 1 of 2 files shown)


src/test/java/tests/ConformanceTest.java
```diff
@@ -129,9 +130,16 @@ protected boolean mustBeExpected() {
...
+      switch (detailMessage.messageKey) {
+        case "sourceType":
+          return expressionType(
+              fixType(detailMessage.messageArguments.get(0)),
+              detailMessage.messageArguments.get(1));
+
+        case "sinkType":
+          return sinkType(
+              fixType(detailMessage.messageArguments.get(0)),
+              detailMessage.messageArguments.get(1).replaceFirst("^[^.]+\\.", ""));
```
cpovirk: Is this stripping a class and package name? It's probably worth a comment.

Also, would it make sense to do that work inside the checker, rather than when the test consumes the results?

cpovirk (approved):

src/test/java/tests/ConformanceTest.java
netdpb: It's stripping just the simple name of the class, which is what CF emits by default.

My feeling is that there's a judgment call about where to adapt the checker to the conformance tests. I'm leaning on basic enabling of sufficient diagnostics in the checker, and string massaging in the adapter.

cpovirk: Mostly it just feel strange to me that we'd output data that forces ourselves to munge it. And maybe hypothetically it would be nice to be able to run the checker directly in this debugging mode and see the same simple output that the conformance tests produce.

But what you have is fine.

PR #108, opened 2023-11-15
Move conformance test framework code into a subproject

netdpb:
Part of #107.


conformance/build.gradle
```diff
@@ -0,0 +1,24 @@
+plugins {
+    id 'java-library'
+}
+
+group 'org.jspecify'
+version '0.1'
```
cpovirk, line 6: I've liked the 0.0.0-SNAPSHOT pattern to simplify cross-project dependencies. Maybe that's not an issue, though, as long as everything involved uses Gradle and uses an `include` of our project? Then again, if that's the case, maybe we should be using an `include` for `jspecify` instead of a dependency on its 0.0.0-SNAPSHOT version? (Or maybe we should pick a "real" release of `jspecify`?)

conformance/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+
+repositories {
+    mavenCentral()
+}
+
+java {
+    sourceCompatibility = 1.9
```
cpovirk, line 13: We might want to bump to 11 at some point (and bump the Guava dep, too). But I see that you're just carrying over what we have in the existing project.

cpovirk (approved):

conformance/build.gradle
netdpb: I was doing what Gradle suggests for now. I think it's more work to set up a Maven artifact, which we will have to do later anyway.

conformance/build.gradle
netdpb: Exactly.

PR #110, opened 2023-11-16
Gradle build cleanup

netdpb:
- Clean up dependency references, including to the conformance test framework.
- Use a version catalog.
- Set the conformance test framework version to "0-SNAPSHOT".
- Clean up Gradle warnings.


conformance/build.gradle
```diff
@@ -3,7 +3,7 @@ plugins {
 }
 
 group 'org.jspecify'
-version '0.1'
+version '0-SNAPSHOT'
```
cpovirk: I went with full-on "0.0.0-SNAPSHOT" for Semantic Versioning reasons, but I am not overly invested in that.

cpovirk (approved):

settings.gradle
```diff
@@ -19,3 +23,25 @@ includeBuild("../jdk") {
 includeBuild("../checker-framework") {
     initializeProject()
 }
+
+dependencyResolutionManagement {
+    versionCatalogs {
+        libs {
+            version("checkerFramework", "3.21.5-SNAPSHOT")
+
+            library("checkerFramework-checker", "org.checkerframework", "checker").versionRef("checkerFramework")
```
wmdietl, line 32: I was confused why here the libraries use `-`, but then in the uses in `build.gradle` the libraries use `.`.
https://docs.gradle.org/current/userguide/platforms.html#sub:version-catalog-declaration explains this and the recommendation (for whatever reason) is to use `-` here, but then requires to use `.` in the uses...
Is it worth adding that manual URL here?

conformance/build.gradle
wmdietl: Being consistent between this and jspecify/jspecify would be nice.

wmdietl (approved):
Thanks, looks great!

conformance/build.gradle
netdpb: OK.

settings.gradle
netdpb: I don't think it is worth it. This is how version catalogs work; it's not a choice we're making. I wouldn't expect any usages of features of Gradle to be documented.

The reason it's dots in  `build.gradle` is because these become objects with properties: `libs` has a property `checkerFramework` which has a property `checker`. That's Java/Groovy property notation, and identifiers can't have hyphens.

PR #113, opened 2023-11-28
Upgrade Gradle to 8.5.

netdpb:


cpovirk (approved):

netdpb:
Trying to figure out why the build is breaking.

netdpb:
Can't break it locally. Trying to add `--stacktrace` here.

netdpb:
I give up.

wmdietl:
@netdpb The CI failure looks like something goes wrong with annotation-tools.
Maybe these two help: https://github.com/jspecify/checker-framework/pull/40 and https://github.com/jspecify/checker-framework/pull/41
It works for me locally :-)

netdpb:
It works for me locally too, even without those.

Yes, it's an issue with annotation-tools, which also uses an old Gradle. I think we should give up on this until you've finished deforking. Then we will be able to depend on a released version of annotation-tools (and eisop), right?

wmdietl:
eisop doesn't depend on annotation-tools. Merging the two checker-framework PRs should unblock this PR and wouldn't make deforking harder.

netdpb:
Oh cool! I commented on some Gradle and shell bits of the first one.

cpovirk (approved):

PR #115, opened 2023-12-05
Consume a single conformance test ZIP

netdpb:
Depends on https://github.com/jspecify/jspecify/pull/422.


cpovirk (approved):

netdpb:
I'm having trouble here. On my local machine, the tests pass. On the GitHub action, they fail because when analyzing `UsesDep.java`, Checker Framework reports:

```
UsesDep.java:21 (type.checking.not.run) NullSpecChecker did not run because of a previous error issued by javac
    public class UsesDep {
           ^
```

This error does not occur on my local machine (running with JDK 11 or JDK 17; GitHub's action uses JDK 17).

However, no previous error appears in the output.

netdpb:
OK, I figured it out.

netdpb:
The problem was that I tried to add system properties with the files extracted from the ZIP in Gradle's configuration phase, which runs while the task graph is being set up. Those files don't appear until the _execution_ phase, which is what actually runs for a task after the tasks it depends on have run.

Setting the system properties in `doBefore` blocks did the trick.

cpovirk (approved):
I helpfully see nothing wrong with this version, just as I helpfully saw nothing wrong with the previous version!

PR #116, opened 2023-12-06
Use EISOP CF sources

wmdietl:
As discussed, this is an attempt to break down the changes into smaller PRs.

The scripts should try to clone `eisop/checker-framework` automatically. If you already have a fork of the CF, it will use that, so you'll need to either update that fork or manually clone the CF first.

After this, `./gradlew test` should work, so everything should compile and a minimal sanity check passes. Neither the conformance tests nor the samples will work as well as before. The goal is to have more PRs in the future that will break less and less.

PR #118, opened 2023-12-08
Use EISOP CF sources

wmdietl:
This is a follow-up to #116.
Instead of merging into the `main` branch and breaking CI, this PR merges into a new `main-eisop` branch. CI for that branch will still fail, but we'll have a clean `main` branch and can have smaller PRs against `main-eisop` until that branch passes CI or is considered good enough.


initialize-project
```diff
@@ -69,4 +73,4 @@ git_clone jspecify --no-single-branch
 
 git_clone jdk --depth 1 --single-branch
 
-git_clone checker-framework
+git_clone checker-framework --depth 1 --single-branch
```
cpovirk: We have a `SHALLOW` environment variable that users can set to enable this. That might be better than hard-coding (which, admittedly, we do for the JDK, but it's truly huge). I don't remember exactly how much thought we gave to all this, though.

cpovirk (approved):

.github/workflows/build.yml
```diff
@@ -26,7 +26,9 @@ jobs:
           arguments: conformanceTest conformanceTestOnSamples
       - name: Check out jspecify/samples-google-prototype
         run: git -C ../jspecify checkout samples-google-prototype
+        if: always()
       - name: Run Samples Tests
         uses: gradle/gradle-build-action@v2
         with:
           arguments: jspecifySamplesTest
+        if: always()
```
netdpb: These seem like a good change to make on the main branch, separately.

initialize-project
```diff
@@ -55,12 +55,16 @@ git_clone() {
...
     if run "${git[@]}" "${forking_org}/${repo}.git" "../${repo}"; then
       return
     fi
   fi
-  run "${git[@]}" "https://github.com/jspecify/${repo}.git" "../${repo}"
+  if [[ "${repo}" == "checker-framework" ]]; then
+    run "${git[@]}" "https://github.com/eisop/checker-framework.git" "../checker-framework"
+  else
+    run "${git[@]}" "https://github.com/jspecify/${repo}.git" "../${repo}"
+  fi
```
netdpb:
```suggestion
  if [[ "${repo}" == checker-framework ]]; then
    forking_org=https://github.com/eisop
  else
    forking_org=https://github.com/jspecify
  fi
  run "${git[@]}" "${forking_org}/${repo}.git" "../${repo}"
```

This way we don't repeat too much.

initialize-project
netdpb: The point of `$SHALLOW` is to allow CIs like GitHub actions to avoid cloning any commits they don't care about. If you're just running tests for CI you don't need to check any past commits out.

I think if someone has their own fork of eisop we wouldn't want them to clone only one branch. I would suggest continuing to rely on `$SHALLOW` (defined externally) for `checker-framework`.

src/main/java/com/google/jspecify/nullness/NullSpecVisitor.java
```diff
@@ -115,10 +116,11 @@ protected String extraArgForReturnTypeError(Tree tree) {
      * the possibly null value (possibly both!). However, this gets tricky: If the branches return
      * `Foo?` and `Foo*`, then we ideally want to emphasize the `Foo?` branch *but*, at least in
      * "strict mode," not altogether ignore the `Foo*` branch.
-     */
+     *
```
netdpb:
```suggestion
      /
```

Maybe leave the slash to mirror what you did on line 106.

netdpb:
Interesting!

```
tests.ConformanceTest > conformanceTests FAILED
    expected:
        # 10 pass; 6 fail; 16 total; 62.5% score
        ...
    but was:
        # 4 pass; 12 fail; 16 total; 25.0% score
```

but

```
tests.ConformanceTest > conformanceTestsOnSamples FAILED
    expected:
        # 75 pass; 498 fail; 573 total; 13.1% score
        ...
    but was:
        # 224 pass; 349 fail; 573 total; 39.1% score
```

Score went up from 13.1% to 39.1%. Woohoo!

.github/workflows/build.yml
wmdietl: I opened #121 for this.

netdpb (approved):

netdpb:
I see a failure of depending on that zip artifact. Do you need to merge `main` in again?

Once that's fixed, you can run the following to generate changes to the conformance test reports to reflect the current behavior:

```
JSPECIFY_CONFORMANCE_TEST_MODE=write ./gradelew conformanceTests
```

wmdietl:
> I see a failure of depending on that zip artifact. Do you need to merge `main` in again?

It looks like this PR is even with `main` and I don't see a failure about a zip file. Can you point me to the failure?
I think I got this error locally when I hadn't updated my local `jspecify` clone.

I do see this failure:

````
> Task :conformanceTests

tests.ConformanceTest > conformanceTests FAILED
    java.nio.file.NoSuchFileException: /home/wdietl/Sync/wmdietl/workspaces/jspecify/jspecify-reference-checker-use-eisop/build/conformanceTests/assertions/org/jspecify/conformance/tests
````

locally and [on CI](https://github.com/jspecify/jspecify-reference-checker/actions/runs/7215214404/job/19659017120?pr=118#step:4:259).

I tried cleaning, updating, and rebuilding everything, but I still get this error.
Did you mean this failure?

netdpb:
Yes, that's the error I mean. That directory should have been created when unzipping.

wmdietl:
> Yes, that's the error I mean. That directory should have been created when unzipping.

This was https://github.com/jspecify/jspecify/issues/436 fixed by https://github.com/jspecify/jspecify/pull/437.

initialize-project
cpovirk: Not really related to this PR, but:

> [Use `--filter=blob:none` instead of `--depth 1`](https://github.com/typetools/jdk/commit/6c6241dc47453a3b1bccda4c7dff4df1795241c1)

I wonder if that's worth investigating someday.

wmdietl: I had a quick look at this post:
https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

It seems to generally prefer the blob-less mode. For CI, where one only has a single build, continuing to use `--depth 1` seems okay. But as this script is for developer set-up, using blob-less might be better.

netdpb: Let's postpone that to a different PR. It's not related to switching to EISOP.

PR #119, opened 2023-12-11
Build and test

netdpb:
Build and test in one step, since build already depends on `conformanceTests`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,16 +14,12 @@ jobs:
         with:
           distribution: temurin
           java-version: 17
-      - name: Build
+      - name: Build and Test
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: build
+          arguments: build conformanceTests
         env:
           SHALLOW: 1
-      - name: Run Conformance Tests
-        uses: gradle/gradle-build-action@v2
-        with:
-          arguments: conformanceTests
       - name: Check out jspecify/samples-google-prototype
         run: git -C ../jspecify checkout samples-google-prototype
       - name: Run Samples Tests


.github/workflows/build.yml
```diff
@@ -14,16 +14,12 @@ jobs:
         with:
           distribution: temurin
           java-version: 17
-      - name: Build
+      - name: Build and Test
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: build
+          arguments: build conformanceTests
```
cpovirk, line 20:
```suggestion
          arguments: conformanceTests
```

Given your point about the dependency that's declared, would it make sense to request just `conformanceTests`? If nothing else, it could be nice to know if that request _stops_ working, especially if it's what we'd tell our (legions of) users to run.

cpovirk (approved):
I have sometimes gotten the impression that people define separate steps for their workflow even when they know that it's not necessary. I have never been clear on why, though. So LGTM.

.github/workflows/build.yml
netdpb: Well, `build` depends on more than just `conformanceTests`, so your suggestion would not run all the verification tasks.

It used to be that the conformance tests, like the samples, were separate from the normal Gradle build lifecycle. But now they're not. `conformanceTests` is part of `check`, which is the lifecycle task for all verification, and which `build` depends on.

cpovirk: Oh, thanks. For starters, I did have it backward: While your message had said that `build` depends on `conformanceTests`, I read it the other way around (presumably from being used to the Bazel terminology).

I guess that my question _should_ have been whether we might want to pass only `build`?

netdpb: I considered that. But I thought this would make it a little more explicit. I checked, and this doesn't run `conformanceTests` twice.

cpovirk: :+1:

PR #121, opened 2023-12-14
Always check samples in CI

wmdietl:
As suggested [here](https://github.com/jspecify/jspecify-reference-checker/pull/118/files#r1423234545) it will be good to see sample results, even if conformance tests might fail.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,10 +21,10 @@ jobs:
         env:
           SHALLOW: 1
       - name: Check out jspecify/samples-google-prototype
-        run: git -C ../jspecify checkout samples-google-prototype
         if: always()
+        run: git -C ../jspecify checkout samples-google-prototype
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -22,7 +22,9 @@ jobs:
           SHALLOW: 1
       - name: Check out jspecify/samples-google-prototype
         run: git -C ../jspecify checkout samples-google-prototype
+        if: always()
```
netdpb, line 24: I think the `if` should come near the top, just after the `name`, since it's a guard condition. (It obviously works correctly no matter where, but I think it's a bit more readable at the top.)

netdpb (approved):

.github/workflows/build.yml
wmdietl: Done.

PR #123, opened 2023-12-19
Use the samples files from the conformance tests zip.

netdpb:
Depends on https://github.com/jspecify/jspecify/pull/438.

Helps to decouple this build from [JSpecify](https://github.com/jspecify/jspecify).


wmdietl (approved):
Thanks, looks good, modulo the CI failure...

netdpb:
> Thanks, looks good, modulo the CI failure...

That should be just because the other PR wasn't merged.

wmdietl (approved):
Much nicer with the updated samples branch and CI passes.

PR #125, opened 2023-12-21
Add a test that `./demo SimpleSample.java` works correctly.

netdpb:
We can update this test as we update the instructions.


cpovirk (approved):
Nice, thanks.

PR #128, opened 2024-01-03
Use recommended pattern for Gradle actions and side-by-side projects.

netdpb:
Set up Gradle in one step and just run it as normal run steps. See [docs](https://github.com/gradle/gradle-build-action?tab=readme-ov-file#use-the-action-to-setup-gradle).

Also check out `jspecify-reference-checker` and `jspecify` side-by-side within the workspace (see [docs](https://github.com/actions/checkout?tab=readme-ov-file#checkout-multiple-repos-side-by-side)), instead of putting `jspecify` and the other siblings _next to_ the workspace.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,25 +6,29 @@ jobs:
   build-and-test:
     name: Build and Test
     runs-on: ubuntu-latest
+    defaults:
+      run:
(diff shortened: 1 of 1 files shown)


wmdietl (approved):
Nice!

PR #130, opened 2024-01-08
Handle unexpected diagnostics that aren't in "detail message" format.

netdpb:
Otherwise they were getting swallowed and never emitted even in details mode.


cpovirk (approved):
> Handle unexpected diagnostics that aren't emitted by the framework as "detail messages".

I mis-parsed that at first: The diagnostics _are_ emitted, but they're emitted in a format other than the detail-message format :) SGTM.

netdpb:
Thanks. I've updated the title to be less misparseable.

cpovirk (approved):

PR #134, opened 2024-01-11
Use the new `TypeInformationPresenter` to output more type information

wmdietl:
Depends on https://github.com/eisop/checker-framework/pull/680

Fixes https://github.com/jspecify/checker-framework/issues/44


src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
```diff
@@ -0,0 +1,107 @@
...
+      String direction = null;
+      String usage = null;
+
+      switch (tree.getKind()) {
+        case RETURN:
+          direction = "sinkType";
+          usage = "return";
+          break;
+
+        case METHOD_INVOCATION:
```
wmdietl, line 83: This part isn't finished. I think for the simple code in `SimpleSample.java` it outputs the same information as `main`. For method calls it doesn't match the format yet. For many other trees it outputs too much or not enough.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
```diff
@@ -0,0 +1,107 @@
...
+import com.sun.source.tree.Tree;
+import java.util.List;
+import org.checkerframework.framework.type.AnnotatedTypeFactory;
+import org.checkerframework.framework.type.AnnotatedTypeFormatter;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedExecutableType;
+import org.checkerframework.framework.util.visualize.AbstractTypeInformationPresenter;
+import org.checkerframework.framework.util.visualize.TypeOccurrenceKind;
+
+public class ConformanceTypeInformationPresenter extends AbstractTypeInformationPresenter {
```
netdpb: Should this be `final`?

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
```diff
@@ -0,0 +1,107 @@
...
+          direction = "sinkType";
+          usage = "return";
+          break;
+
+        case METHOD_INVOCATION:
+          AnnotatedExecutableType calledType = (AnnotatedExecutableType) type;
+          List<? extends AnnotatedTypeMirror> params = calledType.getParameterTypes();
+          MethodInvocationTree mit = (MethodInvocationTree) tree;
+          List<? extends ExpressionTree> args = mit.getArguments();
+          assert params.size() == args.size();
```
netdpb, line 90: Will this be true for variadic methods?

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
```diff
@@ -0,0 +1,107 @@
...
+
+          for (int i = 0; i < params.size(); ++i) {
+            checker.reportError(
+                tree,
+                "sinkType",
+                typeFormatter.format(params.get(i)),
+                // TODO: build up something from method and parameter name?
```
netdpb: What I've done so far is `methodName#paramName`. See [_Basic.java_ line 42](https://github.com/jspecify/jspecify/blob/main/conformance-tests/src/assertions/java/org/jspecify/conformance/tests/Basic.java#L42).

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
```diff
@@ -0,0 +1,107 @@
...
+                "param");
+
+            checker.reportError(
+                tree,
+                "sourceType",
```
netdpb: I think this should just be "expressionType". We use "sourceType" only in a cannot-convert error, at least for now.

I think the framework expects any expression (but maybe not completely unused expressions) to have its type reported as "expressionType". So I would expect this logic to be applicable to any `ExpressionTree`.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
```diff
@@ -0,0 +1,107 @@
...
+                typeFormatter.format(atypeFactory.getAnnotatedType(args.get(i))),
+                // TODO: build up something from method and parameter name?
+                "arg");
+          }
+          break;
+        case IDENTIFIER:
+          if (occurrenceKind == TypeOccurrenceKind.USE_TYPE) {
+            direction = "sourceType";
```
netdpb: Same here: "expressionType".

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
netdpb: Too much is okay.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
```diff
@@ -0,0 +1,107 @@
...
+                typeFormatter.format(params.get(i)),
+                // TODO: build up something from method and parameter name?
+                "param");
+
+            checker.reportError(
+                tree,
+                "sourceType",
+                typeFormatter.format(atypeFactory.getAnnotatedType(args.get(i))),
+                // TODO: build up something from method and parameter name?
+                "arg");
```
netdpb: Like other "expressionType" reports, this should be the text of the expression.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
wmdietl: Sure.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
wmdietl: I had used `reportError` instead of `reportWarning`, which caused the mismatches. This is fixed in the upcoming commit.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
wmdietl: It should be - the var args method should be expanded in `AnnotatedExecutableType` to have the same number of arguments as the tree.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
wmdietl: Done.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
wmdietl: `sourceType` is what I see [here](https://github.com/jspecify/checker-framework/blob/c9cfa7666bb5cc235e737edbb3d466e541ba96a1/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java#L2831) and seems to work better than `expressionType`.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
wmdietl: Done.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
wmdietl: Done.

tests/ConformanceTest-report.txt
```diff
@@ -1,11 +1,11 @@
...
-FAIL: Basic.java:34 test:expression-type:Object!:nonNull
-FAIL: Basic.java:34 test:sink-type:Object?:return
-FAIL: Basic.java:41 test:sink-type:Object?:nullableObject
-FAIL: Basic.java:43 test:sink-type:String!:testSinkType#nonNullString
+PASS: Basic.java:34 test:expression-type:Object!:nonNull
+PASS: Basic.java:34 test:sink-type:Object?:return
+PASS: Basic.java:41 test:sink-type:Object?:nullableObject
+PASS: Basic.java:43 test:sink-type:String!:testSinkType#nonNullString
 FAIL: Basic.java:49 test:expression-type:List!<capture of ? extends String?>:nullableStrings
 FAIL: Basic.java: no unexpected facts
```
wmdietl: This is the last mis-match against `main`. I manually looked into the extra error and I think it is:

````
../jspecify/conformance-tests/src/assertions/java/org/jspecify/conformance/tests/Basic.java:47: error: (type.argument.type.incompatible) $$ 4 $$ E $$ List $$ capture#85 of ? extends String* $$ Object* $$ ( 1525, 1551 ) $$ incompatible type argument for type parameter E of List.
  Object testWildcard(List<? extends @Nullable String> nullableStrings) {
                           ^
  found   : capture#85 of ? extends String*
  required: Object*
````

Here, `Basic.java` failed already before, so that's fine.

However, for `tests/ConformanceTestOnSamples-report.txt` there are now many more failing tests because of `FAIL: AnnotatedReceiver.java: no unexpected facts`, which previously passed.
How can the changes in this PR cause more unexpected facts? It's only outputting type information as warnings, which should be ignored.

@netdpb Do you understand why this happens?
How can I output the unexpected facts most conveniently?

File: src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java (hunk 6 above)
Comment (wmdietl) [reply src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java]: Not done, should have been "see above about sourceType vs. expressionType".

File: src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java (hunk 5 above)
Comment (netdpb) [reply src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java]: You are right. That's an interesting inconsistency in the way I was thinking about these. I think that's related to the detail about unused expressions, in that I think it's not unusual for an analyzer to discard all type information for them.

File: tests/ConformanceTest-report.txt (hunk 8 above)
Comment (netdpb) [reply tests/ConformanceTest-report.txt]: That would be any extra errors.

You can see them by [setting the `JSPECIFY_CONFORMANCE_TEST_MODE` environment variable](https://github.com/jspecify/jspecify-reference-checker/blob/ec180c5af659ac9c242672a8a0e84c63090dd195/conformance-test-framework/src/main/java/org/jspecify/conformance/ConformanceTestRunner.java#L157) to "details".

File: src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java [hunk 9]
```diff
@@ -57,50 +60,56 @@ class ConformanceTypeInformationReporter extends TypeInformationReporter {
...
+                args.get(i).toString());
           }
           break;
         default:
+          // Nothing special for other trees.
       }
 
-      if (direction != null && usage != null) {
-        checker.reportError(tree, direction, typeFormatter.format(type), usage);
+      if (TreeUtils.isExpressionTree(tree)) {
```
netdpb, line 103: Do you need the previous "sourceType" reports at all if you're doing this?

wmdietl: Right! In an earlier version I output the argument position too, but that went away, so this seems to be enough for all `sourceType`s.

tests/ConformanceTest-report.txt
wmdietl: When I try that, I see lots of warnings about `@NullnessUnspecified`:

````
    OOPS: NotNullMarkedAnnotatedWildcardUnspec.java:17 () cannot find symbol
    import org.jspecify.annotations.NullnessUnspecified;
                                   ^
      symbol:   class NullnessUnspecified
      location: package org.jspecify.annotations
````

Do you see those on `main`?

Also, there are many messages like this:

````
    Jan. 16, 2024 6:12:47 P.M. tests.ConformanceTest$DetailMessageReportedFact fixType
    WARNING: type "void bar(Test, T)" did not match /(?<raw>[^<,?!*]+)(?:<(?<args>.+)>)?(?<suffix>[?!*])?/
````

Maybe these are because I output types for methods and that pattern doesn't work for that?
I can filter out method invocations to avoid those, but there are also similar messages about wildcards on `main`.

Why do we need to fix up types? Couldn't we use a `TypeFormatter` that outputs the information in the needed format, by (optionally) modifying https://github.com/jspecify/jspecify-reference-checker/blob/ec180c5af659ac9c242672a8a0e84c63090dd195/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java#L1542 ?

When trying to compare against the output on `main`, I noticed #135.

Comment (netdpb) [reply tests/ConformanceTest-report.txt]: Yes, we see those `NullnessUnspecified` errors on main. Hence https://github.com/jspecify/jspecify/issues/361. That's one of the differences between main and samples-google-prototype, which (still) has that annotation.

You can ignore the warnings about fixing types. Those are there to remind us that the [`fixType`](https://github.com/jspecify/jspecify-reference-checker/blob/ec180c5af659ac9c242672a8a0e84c63090dd195/src/test/java/tests/ConformanceTest.java#L243) method is incomplete. I don't know whether you should emit types for entire methods. But unless you're doing extra work to emit those, I think it would be better to filter in the conformance test itself. In any case, they're just warnings.

We could emit the types in the format that the JSpecify tests expect, but I thought it would be better to put test-specific logic into the test, rather than into the analysis tool itself. We can discuss that decision.

Comment (wmdietl) [reply tests/ConformanceTest-report.txt]: Ah, that makes sense. Let's discuss the type representation separately.

I think many of the new failures are caused by these compilation failures.
I do not see why these compilation issues only arise with this PR and not before on `main-eisop`.
The numbers go from 75 samples passing on `main` to 67 samples passing with this PR, which seems okay.
So I think this PR fixes the source/sink type output.

**Comment by wmdietl** [comment]
This PR will close https://github.com/jspecify/checker-framework/issues/44.

File: src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java [hunk 10]
```diff
@@ -0,0 +1,106 @@
...
+import javax.lang.model.element.ExecutableElement;
+import org.checkerframework.framework.type.AnnotatedTypeFactory;
+import org.checkerframework.framework.type.AnnotatedTypeFormatter;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedExecutableType;
+import org.checkerframework.framework.util.visualize.AbstractTypeInformationPresenter;
+import org.checkerframework.framework.util.visualize.TypeOccurrenceKind;
+import org.checkerframework.javacutil.TreeUtils;
+
+public final class ConformanceTypeInformationPresenter extends AbstractTypeInformationPresenter {
```
netdpb, line 36: I like to see Javadoc for all public classes, if possible.

tests/ConformanceTestOnSamples-report.txt
```diff
@@ -1,16 +1,16 @@
-# 224 pass; 349 fail; 573 total; 39.1% score
+# 67 pass; 506 fail; 573 total; 11.7% score
```
netdpb: What's your guess as to why we see new unexpected facts? Or do we need to fix https://github.com/jspecify/jspecify-reference-checker/issues/135 to find out?

netdpb (approved):

tests/ConformanceTestOnSamples-report.txt
netdpb: See https://github.com/jspecify/jspecify-reference-checker/pull/136 for that fix.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
wmdietl: I've added some now.

src/main/java/com/google/jspecify/nullness/ConformanceTypeInformationPresenter.java
```diff
@@ -0,0 +1,110 @@
...
+        Tree tree, AnnotatedTypeMirror type, TypeOccurrenceKind occurrenceKind) {
+      switch (tree.getKind()) {
+        case ASSIGNMENT:
+          AssignmentTree asgn = (AssignmentTree) tree;
+          AnnotatedTypeMirror varType;
+          if (genFactory != null) {
+            varType = genFactory.getAnnotatedTypeLhs(asgn.getVariable());
+          } else {
+            varType = atypeFactory.getAnnotatedType(asgn.getVariable());
+          }
```
netdpb: Nit: would a ternary be easier to read here? You'd be able to initialize on declaration.

src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
```diff
@@ -305,6 +307,12 @@ private NullSpecAnnotatedTypeFactory(
       withMostConvenientWorld = this;
     }
 
+    if (checker.hasOption("showTypes")) {
```
netdpb: Same nit about ternary. Feel free to ignore.

netdpb (approved):

PR #136, opened 2024-01-19
Handle the `dereference` key correctly for legacy samples tests.

### LINKED ISSUE #135: Detailed conformance test run fails
Running `./gradlew conformanceTests` passes as expected.

Running in detailed mode via `JSPECIFY_CONFORMANCE_TEST_MODE=details ./gradlew conformanceTests` however fails with:

````
    Jan. 16, 2024 6:33:19 P.M. tests.ConformanceTest$DetailMessageReportedFact fixType
    WARNING: type "capture#549 of ? extends Object" did not match /(?<raw>[^<,?!*]+)(?:<(?<args>.+)>)?(?<suffix>[?!*])?/
    Jan. 16, 2024 6:33:19 P.M. tests.ConformanceTest$DetailMessageReportedFact fixType
    WARNING: type "capture#885 of ? extends Object" did not match /(?<raw>[^<,?!*]+)(?:<(?<args>.+)>)?(?<suffix>[?!*])?/

tests.ConformanceTest > conformanceTestsOnSamples FAILED
    java.lang.IndexOutOfBoundsException: index (1) must be less than size (1)

2 tests completed, 1 failed
````

I would expect additional output, not an IOOBE.

netdpb:
It has only one argument, so it doesn't fit the `cannot-convert` assertion format. Later we will add explicit support for it, possibly with a new assertion type, like `dereference-nullable`.

Instead of throwing on poorly formatted diagnostics, just leave them as unexpected. This fixes #135.

### DIFF
--- a/src/test/java/tests/ConformanceTest.java
+++ b/src/test/java/tests/ConformanceTest.java
@@ -157,13 +157,14 @@ public final class ConformanceTest {
   /** A {@link ReportedFact} parsed from a Checker Framework {@link DetailMessage}. */
   static final class DetailMessageReportedFact extends ReportedFact {
 
-    private static final ImmutableSet<String> NULLNESS_MISMATCH_KEYS =
+    private static final String DEREFERENCE = "dereference";
+
+    private static final ImmutableSet<String> CANNOT_CONVERT_KEYS =
         ImmutableSet.of(
             "argument",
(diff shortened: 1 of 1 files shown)


wmdietl (approved):
Makes sense.

PR #139, opened 2024-01-22
Fix the root project version so the Nexus Publish plugin knows to publish snapshots

netdpb:
It's weird that it requires this at the root only. This should fix the build.


cpovirk (approved):
Yeah, that is not what I would have guessed, either.

netdpb:
Had to fix the demo script as well.

cpovirk (approved):

netdpb:
This fixes #107.

Issue #141, opened 2024-01-24
Support naming conformance test assertions


netdpb:
Conformance tests should be able to have a name, such that the combination of file name, test name, and specific assertion (like "cannot-convert:Object? to Object!") is unique.

That allows tests to be better documented, and allows us to stop relying on the line number to disambiguate in reports, which is brittle to unrelated changes to the test file.

It is important that discovering the name not require parsing of the Java syntax, but be entirely textual.

The test framework would assert that the combination of test name and assertion be unique within each file for assertions associated with a test name.

Any assertion not associated with a test name would report the line number of the line being tested, as it does today.

PR #143, opened 2024-01-29
Refactoring in preparation of adding support for test names

netdpb:
Extracted `Fact`. `ExpectedFact`, and `ReportedFact` to top level types, and adjusted their accessibility.

Added `ExpectedFact.Reader` to contain logic to read expected facts from files, and `ConformanceTestReport.Builder` to build up a report.

Added unit tests for `ExpectedFact.Reader`.

Preparation for fixing #141.


conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java
```diff
@@ -0,0 +1,167 @@
...
+
+    private static final Pattern EXPECTATION_COMMENT =
+        Pattern.compile(
+            "\\s*// "
+                + ("(?<expectation>"
+                    + ASSERTION_PATTERNS.stream().map(Pattern::pattern).collect(joining("|"))
+                    + ")")
+                + "\\s*");
+
+    private final Map<Long, String> facts = new HashMap<>();
```
wmdietl: These two are fields to avoid the allocation of a new `HashMap` for multiple calls of `readExpectedFacts` on the same `Reader`?

conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java
```diff
@@ -0,0 +1,167 @@
...
+        String line = i.next();
+        lineNumber = i.nextIndex();
+        Matcher matcher = EXPECTATION_COMMENT.matcher(line);
+        if (matcher.matches()) {
+          String expectation = matcher.group("expectation");
+          if (expectation != null) {
+            facts.put(lineNumber, expectation.trim());
+          }
+        } else {
+          facts.forEach(
```
wmdietl, line 132: This assumes that the last line is not matching `EXPECTATION_COMMENT`, right? Otherwise, `facts` might contain things that will never be added to `expectedFacts.
A comment might help here, or an assertion before the `return` that `facts` is empty.

conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java
```diff
@@ -0,0 +1,167 @@
+// Copyright 2023 The JSpecify Authors
```
wmdietl, line 1: Should a new file use `2024` or is it more important that some of the logic in this file existed in 2023 already?

conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java
```diff
@@ -0,0 +1,167 @@
...
+  private final long factLineNumber;
+
+  ExpectedFact(Path file, long lineNumber, String fact, long factLineNumber) {
+    super(file, lineNumber);
+    this.fact = fact;
+    this.factLineNumber = factLineNumber;
+  }
+
+  /**
+   * Returns an expected fact representing that the source type cannot be converted to the sink type
```
wmdietl: From this description I expected the return type to be `ExpectedFact`. Is there a better term for the returned String?

wmdietl (commented):
I have a few questions.

conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java
netdpb: This is mostly just extracting code from one file into another. so I kept the copyright year. I see guidance that says not to change the copyright year when moving or copying a file. You could consider this copying-then-editing.

But if someone has more specific guidance, I'm happy to follow it.

conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java
netdpb: Yes. Based on this I've tried to use "factText" throughout, and merged `ReportedFact.expectedFact()` into `ReportedFact.getFactText()`.

I also noticed that these static methods are only used from implementations of `ReportedFact`, so I moved them there.

conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java
netdpb: Sort of. they don't need to be fields yet, but it'll help when I add support for test names. I'll remove them from this.

conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java
netdpb: I've added a TODO. The upcoming PR to deal with test names will include error handling.

netdpb (commented):
Thanks.

netdpb:
You can see a preview of how I'm going to use this refactoring in https://github.com/netdpb/reference-checker/pull/10.

wmdietl (approved):
Thanks, looks good!

PR #144, opened 2024-01-30
Only handle messages with details

wmdietl:
#130 fixed handling of unexpected messages.
However, only the [test suite filtered out such messages](https://github.com/jspecify/jspecify-reference-checker/pull/130/files#diff-d0ccedbf8ecabed5c350d62c5e7831452fb612309cba05da83f0c13d2f928eb0R109).
This makes the corresponding change to the conformance test suite.

Before this PR, the conformance tests in main-eisop failed unexpectedly, as the main-eisop branch was rebased with the changes from #130 and the failure wasn't noticed until #134.

PR #145, opened 2024-01-30
Only handle messages with details and don't expect `type.checking.not.run` errors

wmdietl:
Based on what I did in #144, but here we also need to filter out the `type.checking.not.run` errors.
These errors were removed in more recent CF versions: https://github.com/eisop/checker-framework/blob/master/docs/CHANGELOG.md#version-3350-june-1-2023

If we merge this, we should remove the `type.checking.not.run` logic in #144.

### DIFF
--- a/src/test/java/tests/ConformanceTest.java
+++ b/src/test/java/tests/ConformanceTest.java
@@ -147,6 +147,7 @@ public final class ConformanceTest {
     return result.getUnexpectedDiagnostics().stream()
         .map(d -> DetailMessage.parse(d.getMessage(), testDirectory))
         .filter(Objects::nonNull)
+        .filter(DetailMessage::hasDetails)
         .map(DetailMessageReportedFact::new)
         .collect(toImmutableSet());
   }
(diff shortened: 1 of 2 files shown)


wmdietl:
`PASS: AnnotatedReceiver.java: no unexpected facts` is rather misleading: that test fails because `@NullnessUnknown` isn't found.
But because we filter out javac errors and now also `type.checking.not.run` errors, the test passes, because there were no expected facts.
I'm not sure whether we should merge this and similarly #144.

wmdietl:
#146 is probably better for `main-eisop` and we can drop this PR. Let me know if I should make a PR that documents that we don't filter here, like in #146.

netdpb:
I also prefer #146 to this. If there are no expectations in a file, and a tool "passes" only because it bailed out of analyzing the file at all, that shouldn't count.

(Of course, with the more detailed conformance test assertion style, the chance of a test file without assertions goes way down.)

wmdietl:
I agree.

PR #146, opened 2024-01-30
Document that detail messages should not be filtered out

wmdietl:
Alternative to #144. This documents that messages without details should not be filtered out and updates the conformance test report.
This will then reduce the conformance test differences in #134.


netdpb:
Can you paste the results from running with `JSPECIFY_CONFORMANCE_TEST_MODE=details`?

wmdietl:
Output with details attached.

[jrc-details.txt](https://github.com/jspecify/jspecify-reference-checker/files/14119124/jrc-details.txt)

netdpb (approved):

PR #148, opened 2024-02-01
Support test names by comment

### LINKED ISSUE #141: Support naming conformance test assertions

netdpb:
Partially fixes #141.


netdpb:
Ping for review.

conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java
```diff
@@ -108,35 +124,91 @@ static class Reader {
...
+      return checkUniqueTestNames(expectedFacts.build());
+    }
+
+    private void setTestName(@Nullable String testName) {
+      if (testName == null) {
+        return;
+      }
+      check(this.testName == null, "test name already set");
+      check(facts.isEmpty(), "test name must come before assertions for a line");
+      this.testName = testName.trim();
```
wmdietl: Should this first check that the test name isn't empty?

wmdietl (approved):
One possible corner case to look at, otherwise looks good.

conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java
netdpb: Good catch!

PR #149, opened 2024-02-03
Try to fix dependabot configuration

wmdietl:
Spotless isn't being updated from [6.14](https://github.com/jspecify/jspecify-reference-checker/blob/17cd137c8a0fd8723d892be3a4ff2ffac551be46/build.gradle#L9) to the current 6.25 and several other versions are also out of date.
Maybe adding quotes helps.


cpovirk (approved):
YAML....

PR #153, opened 2024-02-06
Report irrelevant `@NonNull` annotations

netdpb:
Wait until #152 is merged.

### DIFF
--- a/src/main/java/com/google/jspecify/nullness/NullSpecVisitor.java
+++ b/src/main/java/com/google/jspecify/nullness/NullSpecVisitor.java
@@ -28,6 +28,7 @@ import static java.util.Arrays.asList;
 import static javax.lang.model.element.ElementKind.ENUM_CONSTANT;
 import static javax.lang.model.element.ElementKind.PACKAGE;
 import static org.checkerframework.framework.util.AnnotatedTypes.asSuper;
+import static org.checkerframework.javacutil.AnnotationUtils.annotationName;
 import static org.checkerframework.javacutil.AnnotationUtils.areSameByName;
 import static org.checkerframework.javacutil.TreeUtils.annotationsFromTypeAnnotationTrees;
 import static org.checkerframework.javacutil.TreeUtils.elementFromDeclaration;
@@ -568,6 +569,15 @@ final class NullSpecVisitor extends BaseTypeVisitor<NullSpecAnnotatedTypeFactory
     }
   }
 
+  private static final Set<String> NULLNESS_ANNOTATIONS =
+      Set.of(
+          "org.jspecify.annotations.NonNull",
+          "org.jspecify.annotations.Nullable",
+          "org.jspecify.annotations.NullnessUnspecified",
+          "org.jspecify.nullness.NonNull",
+          "org.jspecify.nullness.Nullable",
+          "org.jspecify.nullness.NullnessUnspecified");
+
   private void checkNoNullnessAnnotations(
       Tree treeToReportOn, List<? extends AnnotationTree> annotations, String messageKey) {
     for (AnnotationMirror annotation : annotationsFromTypeAnnotationTrees(annotations)) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/jspecify/nullness/NullSpecVisitor.java
```diff
@@ -579,11 +589,8 @@ private void checkNoNullnessAnnotations(
        * represent the types internally in AnnotatedTypeMirror instances. Contrast this to almost
        * all other logic in the checker, which operates on the internal types.
        */
-      if (areSameByName(annotation, "org.jspecify.annotations.Nullable")
-          || areSameByName(annotation, "org.jspecify.annotations.NullnessUnspecified")
-          || areSameByName(annotation, "org.jspecify.nullness.Nullable")
-          || areSameByName(annotation, "org.jspecify.nullness.NullnessUnspecified")) {
-        checker.reportError(treeToReportOn, messageKey);
+      if (NULLNESS_ANNOTATIONS.stream().anyMatch(na -> areSameByName(annotation, na))) {
```
wmdietl, line 583: Could some of that TODO comment be cleaned up?

src/test/java/tests/ConformanceTest.java
```diff
@@ -209,7 +209,12 @@ protected String getFactText() {
         return cannotConvert(sourceType, sinkType);
       }
       if (IRRELEVANT_ANNOTATION_KEYS.contains(detailMessage.messageKey)) {
-        return irrelevantAnnotation("Nullable"); // TODO(dpb): Support other annotations.
+        if (detailMessage.messageArguments.isEmpty()) {
+          // arguments must start with the annotation
+          return toString();
+        }
+        return irrelevantAnnotation(
+            detailMessage.messageArguments.get(0).replaceFirst(".*\\.", ""));
```
wmdietl, line 218: What is this replacing?

src/main/java/com/google/jspecify/nullness/NullSpecVisitor.java
netdpb: I don't think so. It's all still relevant; I just moved the list of annotations to a constant (and added ~one~two).

src/test/java/tests/ConformanceTest.java
netdpb: It's removing the package name and leaving just the simple name.

wmdietl: I would probably move that logic up to https://github.com/jspecify/jspecify-reference-checker/pull/153/files#diff-5a6f488cb86bfeab841018d6edeaae15b43c2ea77a51a78dda5f398ac92b2fc2R584, so it's together with where the name of the annotation is determined.
I couldn't find a simpler way to get the un-qualified name of one `AnnotationMirror`, just a helper for a set: https://github.com/jspecify/checker-framework/blob/b84882d9b2f77c99081118cc7ad095dfcc3484f5/javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java#L1545

(edit: there is such a helper now: https://github.com/eisop/checker-framework/blob/fd74b33e3db75d3e391436a72fc3d0378dbbdeaf/javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java#L1527)

wmdietl (approved):
One suggestion for you to consider.

src/test/java/tests/ConformanceTest.java
netdpb: Hmm. Are there other places, either in this plugin or in the framework itself, where an annotation's name is included in an error or warning? Where else is an error reported when an annotation is applied incorrectly? I think this plugin should follow whatever the existing practice is.

I think the fully-qualified name is useful to report in case there is more than one annotation with the same simple name in scope.

For types that are reported, it appears that simple names are reported, so you get `List<Foo>` and not `java.util.List<some.pkg.Foo>`. I had assumed that it was not reporting simple names _per se_, but whatever name is in scope.

I don't want to perturb the checker too much just to make the output match the conformance test comment. I think there's a valid reason to do some of that in the test integration. And who knows; maybe we will decide to make the test assertions use the full name of the annotation. If we do it this way, that means we only have to change the test integration and not the checker itself.

PR #154, opened 2024-02-06
Use standard error format

wmdietl:
In https://github.com/eisop/checker-framework/pull/693 I removed some hacky workaround, not realizing that it is still used here.
As we're not planning to write many such tests, I think it's okay to just use the standard error format here.

Ideally, we would have a mechanism to allow type systems to more easily adapt the expected error lines themselves. 
However, at the moment there is a bunch of static methods that make extension hard.

### DIFF
--- a/tests/ConformanceTest-report.txt
+++ b/tests/ConformanceTest-report.txt
@@ -1,4 +1,4 @@
-# 5 pass; 14 fail; 19 total; 26.3% score
+# 46 pass; 68 fail; 114 total; 40.4% score
 FAIL: Basic.java:28:test:expression-type:Object?:nullable
 FAIL: Basic.java:28:test:sink-type:Object!:return
 PASS: Basic.java:28:test:cannot-convert:Object? to Object!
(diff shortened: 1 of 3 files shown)


wmdietl:
@cpovirk The `samples-google-prototype` also uses the special `// jspecify` format. Would you mind if I rewrote that as `// :: error: jspecify`? Hopefully, a simple `sed` script will be able to do that.

wmdietl:
@netdpb It took me a while to figure out why the conformance tests were not updating for me locally. Is it possible that you need to make a new release of `org.jspecify.conformance:conformance-tests`? I only managed to get the new tests when using a local clone and `--include-build`, which shouldn't be a requirement, right?

netdpb:
If you change the samples to use the new format, does the build succeed? If so, I have no objection.

On the other hand, I think you'll need to change `ExpectedFact.NULLNESS_MISMATCH` as well.

netdpb:
> @netdpb It took me a while to figure out why the conformance tests were not updating for me locally. Is it possible that you need to make a new release of `org.jspecify.conformance:conformance-tests`? I only managed to get the new tests when using a local clone and `--include-build`, which shouldn't be a requirement, right?

You may have to run with [`--refresh-dependencies`](https://docs.gradle.org/current/userguide/dynamic_versions.html#sec:refreshing-dependencies) in order to get the latest copy of the tests. Normally Gradle caches snapshot versions for 24 hours.

wmdietl:
> > @netdpb It took me a while to figure out why the conformance tests were not updating for me locally. Is it possible that you need to make a new release of `org.jspecify.conformance:conformance-tests`? I only managed to get the new tests when using a local clone and `--include-build`, which shouldn't be a requirement, right?
> 
> You may have to run with [`--refresh-dependencies`](https://docs.gradle.org/current/userguide/dynamic_versions.html#sec:refreshing-dependencies) in order to get the latest copy of the tests. Normally Gradle caches snapshot versions for 24 hours.

I just ran `./gradlew --refresh-dependencies`.
Afterwards, I still see different behavior between  using `--include-build` or not, even though my local `jspecify` clone is up-to-date.

wmdietl:
> On the other hand, I think you'll need to change `ExpectedFact.NULLNESS_MISMATCH` as well.

Isn't `ExpectedFact` just for the conformance test framework?
Are the files in `samples-google-prototype` processed as CF test cases or as conformance test files?

wmdietl:
I've created a new jspecify branch: https://github.com/jspecify/jspecify/tree/samples-google-prototype-eisop
And [this commit](https://github.com/jspecify/jspecify-reference-checker/pull/154/commits/ea97982d4cabfd5ecebcf91497c9ee3783e086a0) starts using it.
If you look at the CI output before, you'll notice that none of the error markers for the `jspecifySamplesTest` target match.
With this new branch, the errors match again - they don't all pass, but we're back to many expected errors being found.

netdpb:
> > On the other hand, I think you'll need to change `ExpectedFact.NULLNESS_MISMATCH` as well.
> 
> Isn't `ExpectedFact` just for the conformance test framework? Are the files in `samples-google-prototype` processed as CF test cases or as conformance test files?

Both. `jspecifySamplesTest` (which runs [`NullSpecTest`](https://github.com/jspecify/jspecify-reference-checker/blob/main/src/test/java/tests/NullSpecTest.java)) uses the CF test framework; `conformanceTests` (which runs [`ConformanceTest.conformanceTestOnSamples()`](https://github.com/jspecify/jspecify-reference-checker/blob/9e57ea86134fc5de17da55d0ec02ff1fded36df0/src/test/java/tests/ConformanceTest.java#L106)) runs the conformance test framework on the samples. That's why [`ExpectedFact.NULLNESS_MISMATCH`](https://github.com/jspecify/jspecify-reference-checker/blob/9e57ea86134fc5de17da55d0ec02ff1fded36df0/conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java#L44) is there.

netdpb:
> > > @netdpb It took me a while to figure out why the conformance tests were not updating for me locally. Is it possible that you need to make a new release of `org.jspecify.conformance:conformance-tests`? I only managed to get the new tests when using a local clone and `--include-build`, which shouldn't be a requirement, right?
> > 
> > 
> > You may have to run with [`--refresh-dependencies`](https://docs.gradle.org/current/userguide/dynamic_versions.html#sec:refreshing-dependencies) in order to get the latest copy of the tests. Normally Gradle caches snapshot versions for 24 hours.
> 
> I just ran `./gradlew --refresh-dependencies`. Afterwards, I still see different behavior between using `--include-build` or not, even though my local `jspecify` clone is up-to-date.

Which task did you run with `--refresh-dependencies`? You might have to `clean` when switching between including the jspecify build and not.

tests/ConformanceTest-report.txt
netdpb, line 1: It would be nice to have a separate (previously merged) PR just to update the conformance test report, so this PR just has difference from that.

netdpb (approved):

tests/ConformanceTest-report.txt
wmdietl: I split that off into #157 and now this PR is just the changes to adapt the error format.

wmdietl:
Conformance tests and minimal tests pass now.

PR #157, opened 2024-02-07
Update `tests/ConformanceTest-report.txt`

wmdietl:
Update after pulling in `main` into `main-eisop`.


wmdietl:
The `conformanceTests` don't actually run on CI here, because minimal test fails first. I manually checked and with this change the `conformanceTests` pass.

Issue #156, opened 2024-02-07
build instructions for main-eisop


cpovirk:
We probably don't need anything officially published, but for those of us who are looking into the branch now, it would be good to have instructions somewhere.

Here's what I originally tried:

```
$ git clone git@github.com:jspecify/jspecify-reference-checker.git && cd jspecify-reference-checker && git checkout main-eisop && ./gradlew assemble
...
git clone git@github.com:jspecify/checker-framework.git ../checker-framework 
Cloning into '../checker-framework'...

FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':compileJava'.
> Could not resolve all task dependencies for configuration ':compileClasspath'.
   > Could not find io.github.eisop:checker:3.42.0-eisop2-SNAPSHOT.
     Searched in the following locations:
       - file:/usr/local/google/home/cpovirk/.m2/repository/io/github/eisop/checker/3.42.0-eisop2-SNAPSHOT/maven-metadata.xml
       - file:/usr/local/google/home/cpovirk/.m2/repository/io/github/eisop/checker/3.42.0-eisop2-SNAPSHOT/checker-3.42.0-eisop2-SNAPSHOT.pom
       - https://s01.oss.sonatype.org/content/repositories/snapshots/io/github/eisop/checker/3.42.0-eisop2-SNAPSHOT/maven-metadata.xml
       - https://s01.oss.sonatype.org/content/repositories/snapshots/io/github/eisop/checker/3.42.0-eisop2-SNAPSHOT/checker-3.42.0-eisop2-SNAPSHOT.pom
       - https://repo.maven.apache.org/maven2/io/github/eisop/checker/3.42.0-eisop2-SNAPSHOT/maven-metadata.xml
       - https://repo.maven.apache.org/maven2/io/github/eisop/checker/3.42.0-eisop2-SNAPSHOT/checker-3.42.0-eisop2-SNAPSHOT.pom
     Required by:
         project :
   > Could not find io.github.eisop:checker-qual:3.42.0-eisop2-SNAPSHOT.
...
```

I took a couple guesses and discovered that I should go into `checker-framework` and run `git remote add eisop https://github.com/eisop/checker-framework && git fetch eisop master && git checkout eisop/master` (or some similar command of my preference).

For my current testing purposes, I also switched to use the JSpecify JDK that uses the _old_ package name:

[...]

PR #158, opened 2024-02-08
Handle both `https` and `git@` clones

wmdietl:
This should fix another problem reported in #156 .
Is there a better way to generalize this check?
@cpovirk can you try?

### DIFF
--- a/initialize-project
+++ b/initialize-project
@@ -57,7 +57,7 @@ git_clone() {
   forking_org="$(forking_org)"
   if [[ -n "${forking_org}" ]] \
     && [[ "${forking_org}" != "https://github.com/jspecify" ]] \
-    && [[ "${forking_org}" != "git@github.com:/jspecify" ]] ; then
+    && [[ "${forking_org}" != "git@github.com:jspecify" ]] ; then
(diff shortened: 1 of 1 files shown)


initialize-project
```diff
@@ -55,12 +55,14 @@ git_clone() {
 
   local forking_org
   forking_org="$(forking_org)"
-  if [[ -n "${forking_org}" ]] && [[ "${forking_org}" != "https://github.com/jspecify" ]]; then
+  if [[ -n "${forking_org}" ]] \
+    && [[ "${forking_org}" != "https://github.com/jspecify" ]] \
+    && [[ "${forking_org}" != "git@github.com:/jspecify" ]] ; then
```
cpovirk:
```suggestion
    && [[ "${forking_org}" != "git@github.com:jspecify" ]] ; then
```

cpovirk (approved):
Aha, thanks!

I'm still seeing eisop cloned with this PR but only because of a typo :) With the typo fixed, it appears to be working.

initialize-project
wmdietl: Ugh, that's because I've used that URL for my clone... I agree the shorter one is what should be used.

Issue #163, opened 2024-02-08
main-eisop: When `class Foo<T>` is null-unmarked, usages of `T` should have unspecified nullness


cpovirk:
I actually don't know if that's the specific problem here (especially when I'm testing mostly against code that is `@NullMarked`), but it's a good possibility for at least 50 errors \[edit: probably more like 90+?\] that I'm seeing—mostly in https://github.com/google/auto, which uses some unannotated annotation-processor APIs.

```java
import org.jspecify.annotations.NullMarked;

@NullMarked
class NullForUnspecVoid {
  void x(Value val, Visitor<Void> vis) {
    val.accept(vis, null);
  }
}

interface Value {
  <P> void accept(Visitor<P> visitor, P param);
}

interface Visitor<P> {}
```

```
$ ../checker-framework/checker/bin/javac -processorpath $HOME/.m2/repository/org/jspecify/jspecify/0.3.0/jspecify-0.3.0.jar:build/libs/jspecify-reference-checker-0.0.0-SNAPSHOT.jar -processor com.google.jspecify.nullness.NullSpecChecker -AcheckImpl -AassumePure -AsuppressWarnings=contracts.conditional.postcondition.false.methodref,contracts.conditional.postcondition.false.override,contracts.conditional.postcondition.true.methodref,contracts.conditional.postcondition.true.override,purity.methodref,purity.overriding,type.anno.before.decl.anno,type.anno.before.modifier -cp $HOME/.m2/repository/org/jspecify/jspecify/0.3.0/jspecify-0.3.0.jar:build/libs/jspecify-reference-checker-0.0.0-SNAPSHOT.jar NullForUnspecVoid.java
NullForUnspecVoid.java:6: error: [argument.type.incompatible] incompatible argument for parameter param of Value.accept.
    val.accept(vis, null);
                    ^
  found   : null?
  required: Void
1 error
```

I would expect the required type to be `Void*`, since `Visitor` isn't annotated for nullness.

This requires a way to apply defaults to type-variable usages, as discussed in https://github.com/jspecify/checker-framework/issues/7.

wmdietl:
Fixed by #165.

Issue #164, opened 2024-02-08
main-eisop: `x(Foo<? super T>)` cannot override identical signature


cpovirk:
This is responsible for perhaps 90 errors in initial testing on Google's codebase, such as [in `AbstractHasher`](https://github.com/google/guava/blob/5c950ae13a03c8029a1f2cf34704c56d67f4ef76/guava/src/com/google/common/hash/AbstractHasher.java#L131). That _might_ make it the last significant source of errors in that testing. But of course I could be missing others among the noise, and we'll see whether new problems pop up as we fix the old :)

```java
import org.jspecify.annotations.NullMarked;
import org.jspecify.annotations.Nullable;

@NullMarked
interface SuperWildcardParent {
  <T extends @Nullable Object> void x(Foo<? super T> foo);
}

@NullMarked
interface SuperWildcardOverride extends SuperWildcardParent {
  @Override
  <T extends @Nullable Object> void x(Foo<? super T> foo);
}

@NullMarked
interface Foo<T extends @Nullable Object> {}
```

```
$ ../checker-framework/checker/bin/javac -processorpath $HOME/.m2/repository/org/jspecify/jspecify/0.3.0/jspecify-0.3.0.jar:build/libs/jspecify-reference-checker-0.0.0-SNAPSHOT.jar -processor com.google.jspecify.nullness.NullSpecChecker -AcheckImpl -AassumePure -AsuppressWarnings=contracts.conditional.postcondition.false.methodref,contracts.conditional.postcondition.false.override,contracts.conditional.postcondition.true.methodref,contracts.conditional.postcondition.true.override,purity.methodref,purity.overriding,type.anno.before.decl.anno,type.anno.before.modifier -cp $HOME/.m2/repository/org/jspecify/jspecify/0.3.0/jspecify-0.3.0.jar:build/libs/jspecify-reference-checker-0.0.0-SNAPSHOT.jar SuperWildcardOverride.java
SuperWildcardOverride.java:12: error: [override.param.invalid] Incompatible parameter type for foo.
  <T extends @Nullable Object> void x(Foo<? super T> foo);
                                                     ^
  found   : Foo<? super T>
  required: Foo<? super T>
  Consequence: method in SuperWildcardOverride
    void x(Foo<? super T>)
  cannot override method in SuperWildcardParent
    void x(Foo<? super T>)
1 error
```

[...]

PR #165, opened 2024-02-08
Use the new `TypeUseLocation` default locations

### LINKED ISSUE #159: main-eisop crash in TernaryUnspecVsNoQualifier
After building the checker (using the JSpecify JDK that uses the old JSpecify package, for the record, as in https://github.com/jspecify/jspecify-reference-checker/issues/156, but I doubt that's relevant), I cloned `jspecify` as a sibling directory and switched to the `samples-google-prototype` branch. Then:

```
$ ../checker-framework/checker/bin/javac -processorpath ../jspecify/build/libs/jspecify-0.0.0-SNAPSHOT.jar:build/libs/jspecify-reference-checker-0.0.0-SNAPSHOT.jar -processor com.google.jspecify.nullness.NullSpecChecker -AcheckImpl -AassumePure -AsuppressWarnings=contracts.conditional.postcondition.false.methodref,contracts.conditional.postcondition.false.override,contracts.conditional.postcondition.true.methodref,contracts.conditional.postcondition.true.override,purity.methodref,purity.overriding,type.anno.before.decl.anno,type.anno.before.modifier -cp ../jspecify/build/libs/jspecify-0.0.0-SNAPSHOT.jar:build/libs/jspecify-reference-checker-0.0.0-SNAPSHOT.jar ../jspecify/samples/TernaryUnspecVsNoQualifier.java
error: Error in AnnotatedTypeMirror.fromExpression(NullSpecAnnotatedTypeFactory, ImmutableList.of(this)): GenericAnnotatedTypeFactory.addComputedTypeAnnotations:  root needs to be set when used on trees; factory: class com.google.jspecify.nullness.NullSpecAnnotatedTypeFactory
  ; The Checker Framework crashed.  Please report the crash.  Version: Checker Framework 3.42.0-eisop3-SNAPSHOT, branch HEAD, 2024-02-06, commit fd74b33, dirty=true. 
  Checker: class com.google.jspecify.nullness.NullSpecChecker
  Visitor: class com.google.jspecify.nullness.NullSpecVisitor
  Compilation unit: ../jspecify/samples/TernaryUnspecVsNoQualifier.java
  Last visited tree at line 25 column 1:
  class TernaryUnspecVsNoQualifier {
  Exception: java.lang.AssertionError: GenericAnnotatedTypeFactory.addComputedTypeAnnotations:  root needs to be set when used on trees; factory: class com.google.jspecify.nullness.NullSpecAnnotatedTypeFactory; java.lang.AssertionError: GenericAnnotatedTypeFactory.addComputedTypeAnnotations:  root needs to be set when used on trees; factory: class com.google.jspecify.nullness.NullSpecAnnotatedTypeFactory
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.addComputedTypeAnnotations(GenericAnnotatedTypeFactory.java:2025)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.addComputedTypeAnnotations(GenericAnnotatedTypeFactory.java:1991)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.getAnnotatedTypeFromTypeTree(AnnotatedTypeFactory.java:1556)
        at org.checkerframework.framework.type.SupertypeFinder$SupertypeFindingVisitor.supertypesFromTree(SupertypeFinder.java:352)
        at org.checkerframework.framework.type.SupertypeFinder$SupertypeFindingVisitor.visitDeclared(SupertypeFinder.java:191)
        at org.checkerframework.framework.type.SupertypeFinder.directSupertypes(SupertypeFinder.java:63)
        at org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType.directSupertypes(AnnotatedTypeMirror.java:1137)
        at org.checkerframework.framework.util.AnnotatedTypes.getSuperTypes(AnnotatedTypes.java:646)
        at org.checkerframework.framework.util.AnnotatedTypes.overriddenMethods(AnnotatedTypes.java:667)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.inheritOverriddenDeclAnnos(AnnotatedTypeFactory.java:4358)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.getDeclAnnotations(AnnotatedTypeFactory.java:4296)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.getDeclAnnotation(AnnotatedTypeFactory.java:4219)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.getDeclAnnotation(AnnotatedTypeFactory.java:4197)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.getDeclAnnotation(AnnotatedTypeFactory.java:4117)
        at org.checkerframework.framework.util.defaults.QualifierDefaults.defaultsAtDirect(QualifierDefaults.java:765)
        at org.checkerframework.framework.util.defaults.QualifierDefaults.defaultsAt(QualifierDefaults.java:713)
        at org.checkerframework.framework.util.defaults.QualifierDefaults.defaultsAt(QualifierDefaults.java:727)
        at org.checkerframework.framework.util.defaults.QualifierDefaults.applyDefaultsElement(QualifierDefaults.java:855)
        at org.checkerframework.framework.util.defaults.QualifierDefaults.annotate(QualifierDefaults.java:479)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.addComputedTypeAnnotations(GenericAnnotatedTypeFactory.java:2292)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.getAnnotatedType(AnnotatedTypeFactory.java:1441)
        at com.google.jspecify.nullness.NullSpecAnnotatedTypeFactory.getUpperBounds(NullSpecAnnotatedTypeFactory.java:708)
        at com.google.jspecify.nullness.NullSpecAnnotatedTypeFactory.nullnessEstablishingPathExists(NullSpecAnnotatedTypeFactory.java:681)
        at com.google.jspecify.nullness.NullSpecAnnotatedTypeFactory.isNullExclusiveUnderEveryParameterization(NullSpecAnnotatedTypeFactory.java:621)
        at com.google.jspecify.nullness.NullSpecAnnotatedTypeFactory$NullSpecTypeVariableSubstitutor.substituteTypeVariable(NullSpecAnnotatedTypeFactory.java:855)
        at org.checkerframework.framework.type.TypeVariableSubstitutor$Visitor.visitTypeVariable(TypeVariableSubstitutor.java:173)
        at org.checkerframework.framework.type.TypeVariableSubstitutor$Visitor.visitTypeVariable(TypeVariableSubstitutor.java:82)
        at org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedTypeVariable.accept(AnnotatedTypeMirror.java:1861)
        at org.checkerframework.framework.type.AnnotatedTypeCopier.visit(AnnotatedTypeCopier.java:102)
        at org.checkerframework.framework.type.AnnotatedTypeCopier.visitExecutable(AnnotatedTypeCopier.java:202)
        at org.checkerframework.framework.type.AnnotatedTypeCopier.visitExecutable(AnnotatedTypeCopier.java:44)
        at org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedExecutableType.accept(AnnotatedTypeMirror.java:1265)
        at org.checkerframework.framework.type.AnnotatedTypeCopier.visit(AnnotatedTypeCopier.java:95)
        at org.checkerframework.framework.type.TypeVariableSubstitutor.substitute(TypeVariableSubstitutor.java:34)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.methodFromUse(AnnotatedTypeFactory.java:2591)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.methodFromUse(AnnotatedTypeFactory.java:2503)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.methodFromUse(GenericAnnotatedTypeFactory.java:2298)
        at org.checkerframework.framework.type.TypeFromExpressionVisitor.visitMethodInvocation(TypeFromExpressionVisitor.java:413)
        at org.checkerframework.framework.type.TypeFromExpressionVisitor.visitMethodInvocation(TypeFromExpressionVisitor.java:81)
        at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCMethodInvocation.accept(JCTree.java:1666)
        at jdk.compiler/com.sun.source.util.SimpleTreeVisitor.visit(SimpleTreeVisitor.java:80)
        at org.checkerframework.framework.type.TypeFromTree.fromExpression(TypeFromTree.java:42)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.fromExpression(AnnotatedTypeFactory.java:1843)
        at org.checkerframework.framework.type.AnnotatedTypeFactory.getAnnotatedType(AnnotatedTypeFactory.java:1486)
        at org.checkerframework.framework.flow.CFAbstractTransfer.getValueFromFactory(CFAbstractTransfer.java:215)
        at org.checkerframework.framework.flow.CFAbstractTransfer.visitMethodInvocation(CFAbstractTransfer.java:999)
        at com.google.jspecify.nullness.NullSpecTransfer.visitMethodInvocation(NullSpecTransfer.java:152)
        at com.google.jspecify.nullness.NullSpecTransfer.visitMethodInvocation(NullSpecTransfer.java:75)
        at org.checkerframework.dataflow.cfg.node.MethodInvocationNode.accept(MethodInvocationNode.java:124)
        at org.checkerframework.dataflow.analysis.AbstractAnalysis.callTransferFunction(AbstractAnalysis.java:356)
        at org.checkerframework.dataflow.analysis.ForwardAnalysisImpl.callTransferFunction(ForwardAnalysisImpl.java:393)
        at org.checkerframework.dataflow.analysis.ForwardAnalysisImpl.performAnalysisBlock(ForwardAnalysisImpl.java:157)
        at org.checkerframework.dataflow.analysis.ForwardAnalysisImpl.performAnalysis(ForwardAnalysisImpl.java:110)
        at org.checkerframework.framework.flow.CFAbstractAnalysis.performAnalysis(CFAbstractAnalysis.java:150)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.analyze(GenericAnnotatedTypeFactory.java:1623)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.performFlowAnalysis(GenericAnnotatedTypeFactory.java:1513)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.checkAndPerformFlowAnalysis(GenericAnnotatedTypeFactory.java:2100)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.preProcessClassTree(GenericAnnotatedTypeFactory.java:435)
        at com.google.jspecify.nullness.NullSpecAnnotatedTypeFactory.preProcessClassTree(NullSpecAnnotatedTypeFactory.java:1684)
        at org.checkerframework.common.basetype.BaseTypeVisitor.visitClass(BaseTypeVisitor.java:591)
        at org.checkerframework.common.basetype.BaseTypeVisitor.visitClass(BaseTypeVisitor.java:195)
        at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCClassDecl.accept(JCTree.java:808)
        at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:56)
        at org.checkerframework.framework.source.SourceVisitor.visit(SourceVisitor.java:89)
        at org.checkerframework.framework.source.SourceChecker.typeProcess(SourceChecker.java:1136)
        at org.checkerframework.common.basetype.BaseTypeChecker.typeProcess(BaseTypeChecker.java:556)
        at com.google.jspecify.nullness.NullSpecChecker.typeProcess(NullSpecChecker.java:106)
        at org.checkerframework.javacutil.AbstractTypeProcessor$AttributionTaskListener.finished(AbstractTypeProcessor.java:193)
        at jdk.compiler/com.sun.tools.javac.api.ClientCodeWrapper$WrappedTaskListener.finished(ClientCodeWrapper.java:828)
        at jdk.compiler/com.sun.tools.javac.api.MultiTaskListener.finished(MultiTaskListener.java:132)
        at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1415)
        at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1372)
        at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:972)
        at jdk.compiler/com.sun.tools.javac.main.Main.compile(Main.java:311)
        at jdk.compiler/com.sun.tools.javac.main.Main.compile(Main.java:170)
        at jdk.compiler/com.sun.tools.javac.Main.compile(Main.java:57)
        at jdk.compiler/com.sun.tools.javac.Main.main(Main.java:43)
1 error
```

I don't immediately see this crash among my results from running the checker on Google's codebase yester (though it's possible that that's only because we were prevented from getting that far by other errors :)). But I see reference to "root needs to be set when used on trees" in the commits referenced from https://github.com/jspecify/checker-framework/issues/11 (also discussed in https://github.com/jspecify/checker-framework/issues/45).

Given that those posts discuss stub files, and given that I hadn't seen the problem inside Google yet, I'd hoped that the problem would go away if I passed `-Aignorejdkastub`. But I'm seeing it even with that flag.

I do notice `inheritOverriddenDeclAnnos` in the stack trace, so this seems like more reason to provide a way to skip the inheriting of annotations, as discussed in https://github.com/jspecify/checker-framework/issues/11. Indeed, the error goes away if I make a change similar to https://github.com/jspecify/checker-framework/commit/f3e9f21737c5288f601905262f5faf85cfb566d3. (I don't know if the full set of changes is necessary.)

```patch
diff --git a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
index 0dbe8d9f5..34d50389b 100644
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -4268,9 +4268,7 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
 
         AnnotationMirrorSet results = new AnnotationMirrorSet();
         // Retrieving the annotations from the element.
-        // This includes annotations inherited from superclasses, but not superinterfaces or
-        // overridden methods.
-        List<? extends AnnotationMirror> fromEle = elements.getAllAnnotationMirrors(elt);
+        List<? extends AnnotationMirror> fromEle = elt.getAnnotationMirrors();
         for (AnnotationMirror annotation : fromEle) {
             try {
                 results.add(annotation);
@@ -4316,6 +4314,9 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
      */
     private void inheritOverriddenDeclAnnosFromTypeDecl(
             TypeMirror typeMirror, AnnotationMirrorSet results) {
+        if (true) {
+            return;
+        }
         List<? extends TypeMirror> superTypes = types.directSupertypes(typeMirror);
         for (TypeMirror superType : superTypes) {
             TypeElement elt = TypesUtils.getTypeElement(superType);
@@ -4354,6 +4355,9 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
      *     the element itself.
      */
     private void inheritOverriddenDeclAnnos(ExecutableElement elt, AnnotationMirrorSet results) {
+        if (true) {
+            return;
+        }
         Map<AnnotatedDeclaredType, ExecutableElement> overriddenMethods =
                 AnnotatedTypes.overriddenMethods(elements, this, elt);
 
```

### LINKED ISSUE #161: main-eisop: Unbounded wildcards default to non-nullable
This is at least a contributor to perhaps 500ish \[edit: probably even more!\] of the 800ish errors that I've seen in testing inside Google so far:

```java

@NullMarked
class UnboundedDefaultsToNonNull<E extends @Nullable Object> {
  UnboundedDefaultsToNonNull<?> x() {
    return this;
  }
}
```
```
$ ../checker-framework/checker/bin/javac -processorpath $HOME/.m2/repository/org/jspecify/jspecify/0.3.0/jspecify-0.3.0.jar:build/libs/jspecify-reference-checker-0.0.0-SNAPSHOT.jar -processor com.google.jspecify.nullness.NullSpecChecker -AcheckImpl -AassumePure -AsuppressWarnings=contracts.conditional.postcondition.false.methodref,contracts.conditional.postcondition.false.override,contracts.conditional.postcondition.true.methodref,contracts.conditional.postcondition.true.override,purity.methodref,purity.overriding,type.anno.before.decl.anno,type.anno.before.modifier -cp $HOME/.m2/repository/org/jspecify/jspecify/0.3.0/jspecify-0.3.0.jar:build/libs/jspecify-reference-checker-0.0.0-SNAPSHOT.jar UnboundedDefaultsToNonNull.java
UnboundedDefaultsToNonNull.java:7: error: [return.type.incompatible] incompatible types in return.
    return this;
           ^
  type of expression: UnboundedDefaultsToNonNull<E>
  method return type: UnboundedDefaultsToNonNull<? extends Object>
1 error
```

In `@NullMarked` code, `UnboundedDefaultsToNonNull<?>` should be treated like `UnboundedDefaultsToNonNull<? extends @Nullable Object>`, rather than like `UnboundedDefaultsToNonNull<? extends [@NonNull] Object>`. This is https://github.com/jspecify/checker-framework/issues/2.

wmdietl:
Depends on https://github.com/eisop/checker-framework/pull/708.

Something still seems to be missing, failure in `NullUnmarkedUndoesNullMarkedForWildcards.java` is odd.

Fixes #159, #161, #163.


tests/ConformanceTestOnSamples-report.txt
```diff
@@ -70,11 +70,11 @@ FAIL: CaptureConvertedUnspecToOther.java: no unexpected facts
 FAIL: CaptureConvertedUnspecToOtherUnionNull.java: no unexpected facts
 FAIL: CaptureConvertedUnspecToOtherUnspec.java: no unexpected facts
 PASS: CastOfCaptureOfNotNullMarkedUnboundedWildcardForObjectBoundedTypeParameter.java: no unexpected facts
-PASS: CastOfCaptureOfUnboundedWildcardForNotNullMarkedObjectBoundedTypeParameter.java: no unexpected facts
+FAIL: CastOfCaptureOfUnboundedWildcardForNotNullMarkedObjectBoundedTypeParameter.java: no unexpected facts
```
wmdietl, line 73: The unexpected error for [this test case](https://github.com/jspecify/jspecify/blob/d138f92053036b4b82eef42f49ca9b7353230b91/samples/CastOfCaptureOfUnboundedWildcardForNotNullMarkedObjectBoundedTypeParameter.java#L26) is:

````
build/conformanceTests/samples/CastOfCaptureOfUnboundedWildcardForNotNullMarkedObjectBoundedTypeParameter.java:26: error: [type.argument.type.incompatible] incompatible type argument for type parameter T of Supplier.
    void x(Supplier<?> supplier) {
                    ^
  found   : capture#576 of ? extends Object*
  required: Object*
````

I think the defaults work correctly: the `IMPLICIT_WILDCARD_UPPER_BOUND_NO_SUPER` default is `@Nullable`, which gets combined with the declared upper bound, resulting in `capture of ? extends Object*`.

However, that is not a subtype of `Object*`. @cpovirk does this look like some problem with the type hierarchy?

File: tests/minimal/UnboundedDefaultsToNonNull.java [hunk 2]
```diff
@@ -0,0 +1,12 @@
+// Test case for Issue 161:
```
wmdietl: I've added this test case to show that the issue #161 is fixed.
I didn't find a file with this name in the samples.
@cpovirk Should I keep this file here or add it somewhere else?

src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
```diff
@@ -359,7 +396,12 @@ protected void addUncheckedStandardDefaults(QualifierDefaults defs) {
 
   @Override
   protected Set<Class<? extends Annotation>> createSupportedTypeQualifiers() {
-    return new LinkedHashSet<>(asList(Nullable.class, NullnessUnspecified.class, MinusNull.class));
+    return new LinkedHashSet<>(
+        asList(
+            Nullable.class,
+            NullnessUnspecified.class,
+            MinusNull.class,
+            ParametricTypeVariableUse.class));
```
wmdietl: I'll refactor how I do this in EISOP separately and add an explicit qualifier here. This is just a first hack.

tests/ConformanceTestOnSamples-report.txt
cpovirk: It is correct for the checker to produce an error here. The reason that it's showing up as `FAIL` is that the conformance-test framework doesn't support `jspecify_nullness_not_enough_information` yet, [only `jspecify_nullness_mismatch`](https://github.com/jspecify/jspecify-reference-checker/blob/75aa1de2a8f88f8c23347d86465844df9887874c/conformance-test-framework/src/main/java/org/jspecify/conformance/ExpectedFact.java#L45).

(`Object*` is not a subtype of `Object*` so as to indicate that, if you don't know what you have and you don't know what you need, we can't prove that it's safe. Our expectation is that few checkers (if any) will produce even a warning in this case—or even in a more-likely-to-be-unsafe case in which you have an `Object?` and need an `Object*`. Still, the tests label such cases accordingly. This labeling may help tools like the Checker Framework that make specific assumptions about unannotated code: If a line is labeled with `jspecify_nullness_mismatch`, a tool like the Checker Framework can consider itself to "pass" whether it produces an error there or not. This is in contrast to cases in which it should always produce an error and cases in which it never should.)

tests/minimal/UnboundedDefaultsToNonNull.java
cpovirk: Maybe it can live under `conformance-tests/src/assertions`? netdpb@ will be back Thursday :)

tests/ConformanceTestOnSamples-report.txt
wmdietl: Thanks, so I did remember correctly that that is how unspecified nullness is supposed to work.

I've opened https://github.com/jspecify/jspecify/pull/487 to add the optional error.

(In your penultimate sentence: `If a line is labeled with jspecify_nullness_mismatch, a tool like the Checker Framework can consider itself to "pass" whether it produces an error there or not.` Here you meant `jspecify_nullness_not_enough_information`, right?)

cpovirk: Oops, yes, sorry.

wmdietl:
A bit frustratingly, this doesn't fix #163 and #164. At least #163 I had hoped should be fixed by the change to `TYPE_VARIABLE_USE` defaults. I'll look some more.

tests/ConformanceTestOnSamples-report.txt
wmdietl: The error here (which we add as expected in https://github.com/jspecify/jspecify/pull/487) feels wrong to me.
The captured wildcard is for that type parameter of `Supplier`. So while in general, two unspecified nullness can't be subtypes, here, the type argument is a capture of that unspecified nullness, so it will always match.
Even more fundamentally, when checking `Supplier<?>` the wildcard isn't captured at all - a capture only happens at a call/field access. So the found type is not what it should be.
What do you think?

cpovirk: I'm sorry, I was not reading carefully enough :( I had been looking at [line 29](https://github.com/jspecify/jspecify/blob/ec998d519a1716e034a4b1a538d42e9e7ed5c3a6/samples/CastOfCaptureOfUnboundedWildcardForNotNullMarkedObjectBoundedTypeParameter.java#L29), where we do want an error in strict mode.

I agree that line 26 should _not_ be an error even in strict mode.

I will see if I can dig up some commits related to this error, but off the top of my head:

IIRC, it seemed weird to me that the checker was checking whether the captured type was a subtype of the type-parameter bound. I don't recall a similar check in the JLS, perhaps because such a check is necessarily true given how the captured type is constructed :) I wondered if the checker behavior was a holdover from before the checker had an implementation of capture conversion.

In fact, I thought I had some memory that the checker was actually checking whether _the wildcard bound_ was a subtype of the type-parameter bound, which seemed even weirder: Java will let you write `class Foo<T extends CharSequence>` and still write `Foo<? extends Object>`, so why wouldn't the checker let you write `class Foo<T extends Object>` and still write `Foo<? extends @Nullable Object>`? That behavior doesn't seem to be what we're seeing here, since we see the type `capture#576 of ? extends Object*`, which clearly is taking the type-parameter bound into account. So either I'm remembering wrong, or I put in hacks that changed the behavior, or something changed in the checker.

cpovirk: I am still not sure where the code that makes the main branch avoid this error is :\

I thought maybe https://github.com/jspecify/jspecify-reference-checker/commit/89d9b720a64f1b0bb2e6d5add574321f93339625#diff-ce82a59b06d0d82bfc6d06835d1b06ac15619bc3a71617d32189d041f02ca9c0R270, but then I found that I'd removed that cocde in https://github.com/jspecify/jspecify-reference-checker/commit/8ba39b04980f3739eb9c4a0351ee8931683c99a0

I thought maybe https://github.com/jspecify/checker-framework/commit/00cc0561ab759922eba86ff6980787b848cf907a, but then I found that I'd removed that code in https://github.com/jspecify/checker-framework/commit/72c5c16a026fc42b2b076db517b03cab20ed97a4.

Maybe https://github.com/jspecify/jspecify-reference-checker/commit/9348e76115088149d55d9495a5c720a5f1d6cd34#diff-ce82a59b06d0d82bfc6d06835d1b06ac15619bc3a71617d32189d041f02ca9c0R894 is somehow related?

I also have open https://github.com/jspecify/jspecify-reference-checker/commit/74f564f415046f6fc4e0200a7990a0db603db9c8, https://github.com/jspecify/jspecify-reference-checker/commit/7e662e31e4553d81bf4d38cb197b21f54c0fcaca, and https://github.com/jspecify/jspecify-reference-checker/commit/824113a212a7ea001e6557fb800a5d73e8f18232, but I'm not convinced that any of that is relevant.

cpovirk: OK, I started just diffing `checker-framework` against upstream, and the code jumped right out to me. At least, I _assume_ this is the right code:

https://github.com/jspecify/checker-framework/blob/b84882d9b2f77c99081118cc7ad095dfcc3484f5/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java#L562

(That code was a small part of the large merge commit https://github.com/jspecify/checker-framework/commit/2e588e78bd81ee52151bda45720120f3cd83c52a.)

wmdietl: Thanks a lot for tracking this down!
I'll look into fixing this in the reference checker or the framework.
Besides the error here, these incorrect captures are also the cause for #164.

cpovirk: Ah, do you think that it's possible that #164 is related to https://github.com/jspecify/checker-framework/blob/b84882d9b2f77c99081118cc7ad095dfcc3484f5/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java#L5017 (https://github.com/jspecify/checker-framework/issues/47)?

That is, could there be two issues here?
1. The upstream checker performs a check that the wildcard bound (or capture thereof?) is in bounds (of the type-parameter bound? of the wildcard bound?). That leads to failures for JSpecify because `Object*` is not a subtype of `Object*`. My "fix" for that was to remove the possibly unnecessary check entirely. (cause of https://github.com/jspecify/jspecify/pull/487)?)
2. When we do perform capture conversion, we get it wrong. (cause of #164?)

The reason that it seems strange for #164 to be related to this issue is that #164 doesn't involve unspecified nullness.

But I guess that, if the tests of #164 start passing with this change, then it's a fix :) Still, maybe it's more that we're _hiding_ the failure (by no longer checking the incorrect capture against anything) than fixing it (by making the results of capture correct)? I don't know :\

I should also write up something about type parameters in general that I've been thinking about recently. It might end up saving us some work... maybe....

wmdietl: https://github.com/jspecify/jspecify-reference-checker/pull/165/commits/9ee2796e5cd2962f1e8250620528c5b19075d34e uses `hasEffectiveAnnotation` instead of `hasAnnotation`. #164 has `? super T` and when the code asked whether the wildcard had a particular annotation in the lower bound, it got "none", because the type variable `T` isn't annotated. `hasEffectiveAnnotation` follows bounds to find the first declared type.

I replaced all `hasAnnotation` with `hasEffectiveAnnotation` in the `NullSpecAnnotatedTypeFactory`. I should probably have that done more incrementally, as two more conformance tests fail now... let me see why that is. I should also look at uses of `hasAnnotation` in other files.

wmdietl (commented):
@cpovirk Please have a look.
I'll fix #163 separately.

tests/ConformanceTestOnSamples-report.txt
```diff
@@ -303,7 +303,7 @@ FAIL: NullLiteralToTypeVariableUnionNull.java: no unexpected facts
 FAIL: NullLiteralToTypeVariableUnspec.java: no unexpected facts
 FAIL: NullMarkedDirectUseOfNotNullMarkedBoundedTypeVariable.java: no unexpected facts
 PASS: NullUnmarkedUndoesNullMarked.java: no unexpected facts
-PASS: NullUnmarkedUndoesNullMarkedForWildcards.java: no unexpected facts
+FAIL: NullUnmarkedUndoesNullMarkedForWildcards.java: no unexpected facts
```
wmdietl, line 306: [This line](https://github.com/jspecify/jspecify/blob/d138f92053036b4b82eef42f49ca9b7353230b91/samples/NullUnmarkedUndoesNullMarkedForWildcards.java#L24) contains a `jspecify_nullness_not_enough_information` marker.
I get as error:

````
NullUnmarkedUndoesNullMarkedForWildcards.java:25: error: [argument.type.incompatible] incompatible argument for parameter o of NullUnmarkedUndoesNullMarkedForWildcards.acceptObject.
    acceptObject(supplier.get());
                             ^
  found   : Object?
  required: Object
1 error
````

The `Object?` seems wrong, as the wildcard bound should become unspecified.
However, there is an error marker... should I change something about the test case?

**Comment by wmdietl** [comment]
Hmm. The number of expected errors that are not found went to 411.
On the other hand, the number of unexpected errors went to 3.

I can't compare against `main-eisop` on CI, because for some reason it fails there: https://github.com/jspecify/jspecify-reference-checker/actions/runs/8058639754/job/22094984781#step:6:177

````
/home/runner/work/jspecify-reference-checker/jspecify-reference-checker/jspecify-reference-checker/src/main/java/com/google/jspecify/nullness/NullSpecVisitor.java:591: error: cannot find symbol

  private static final Set<String> NULLNESS_ANNOTATIONS =
> Task :compileJava FAILED
                       ^
  symbol:   class Set
  location: class NullSpecVisitor
/home/runner/work/jspecify-reference-checker/jspecify-reference-checker/jspecify-reference-checker/src/main/java/com/google/jspecify/nullness/NullSpecVisitor.java:592: error: cannot find symbol
      Set.of(
      ^
  symbol:   variable Set
  location: class NullSpecVisitor
2 errors
````

What is strange about this is that there is no `Set.of` in that file: https://github.com/jspecify/jspecify-reference-checker/blob/main-eisop/src/main/java/com/google/jspecify/nullness/NullSpecVisitor.java

Also, this PR builds successfully and doesn't touch `NullSpecVisitor.java`. So I'm a bit confused what CI is doing...

**Comment by wmdietl** [comment]
So locally, I get in `main-eisop`:

````
tests.NullSpecTest$Lenient > run[/home/wdietl/tmp/jspecify/jspecify-reference-checker-forked/build/conformanceTests/samples] FAILED
    java.lang.AssertionError: 1203 out of 1225 expected diagnostics were found.
    21 unexpected diagnostics were found:
````

whereas in this branch, as it is right now:

````
tests.NullSpecTest$Lenient > run[/home/wdietl/tmp/jspecify/jspecify-reference-checker-forked/build/conformanceTests/samples] FAILED
    java.lang.AssertionError: 1146 out of 1225 expected diagnostics were found.
    42 unexpected diagnostics were found:
````

So let me see what's going wrong here.

File: tests/ConformanceTestOnSamples-report.txt (hunk 4 above)
Comment (cpovirk) [reply tests/ConformanceTestOnSamples-report.txt:306]: There appears to be a _lot_ going on here.

First: I would have expected:

```
  found   : Object*
  required: Object
```

I think that's what you're saying, as well.

This seems familiar: I think that the reason that `NullUnmarkedUndoesNullMarkedForWildcards` exists is that I had _wanted_ to define `@NullMarked` as something like "`nullnessOperatorUnspecified` for location `ALL`." But that turns out not to work when the enclosing scope is `@NullMarked`: `@NullMarked` configures `unionNull` for `UNBOUNDED_WILDCARD_UPPER_BOUND`, and `UNBOUNDED_WILDCARD_UPPER_BOUND` is more specific than `ALL`, even though it's on a wider scope. Or something like that? The `populateNewDefaults` method in `NullSpecAnnotatedTypeFactory.java` [discusses this](https://github.com/jspecify/jspecify-reference-checker/blob/75aa1de2a8f88f8c23347d86465844df9887874c/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java#L843). Perhaps you just need to do [this TODO](https://github.com/jspecify/jspecify-reference-checker/blob/8eb4ab8280e373c6ec97914a0f1fcf1a62a9e32f/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java#L146)?

Comment (cpovirk) [reply tests/ConformanceTestOnSamples-report.txt:306]: As for the fact that it's an error:

That much is expected. It's only the types that are wrong.

Our checker doesn't actually issue "warnings." It issues only errors. The cases that we would think of as "warnings" are cases where the checker issues an error in strict mode but no error in lenient mode. (That's in contrast to "real" errors, which are errors in both modes.)

This is what I was getting at in https://github.com/jspecify/jspecify/pull/486#issuecomment-1965474896.

As I note there, the conformance-test framework doesn't support `jspecify_nullness_not_enough_information`. Since IIRC the conformance-test framework current runs only in strict mode, it could be made to treat `jspecify_nullness_not_enough_information` exactly like `jspecify_nullness_mismatch`, since both of those comments indicate lines that should be errors in strict mode.

If we were to make the conformance-test framework treat `jspecify_nullness_not_enough_information` in that way, then the test here would begin to pass. It would still be passing _for the wrong reason_ (since it has the types wrong), but we wouldn't notice until we _also_ run the conformance tests in lenient mode. Once we do, it would fail: It would appear there as an unexpected error in lenient mode. Or at least I'd hope so :) (We should be able to see it fail _now_ if we run the old-style samples integration that you'd put into place.)

Comment (cpovirk) [reply tests/ConformanceTestOnSamples-report.txt:306]: But things get more interesting from there....

The current head checker is likely also doing this wrong! I built it and ran `./demo -Astrict` on this test, and:

```
NullUnmarkedUndoesNullMarkedForWildcards.java:25: error: [nullness] possible null dereference on type Supplier<? extends Object*>*
    acceptObject(supplier.get());
                 ^
1 error
```

That's true! Since the method is `@NullUnmarked`, `supplier` has unspecified nullness, so the call to `get()` itself is not provably safe, independent of what we do with the result!

(Hmm, so `main-eisop` should probably be producing _that_ error, too!)

To make that accidental error go away, we should probably change the sample to:

```java
default void apply(@NonNull Supplier<?> supplier) {
```

(Or we could insert a null check in the body of the method.)

But... where's the error that we actually wanted, the one about converting from `Object*` to `Object` in the return statement? It looks like that error is missing at head. I wonder if it ever worked or if we only ever had the null-dereference error, which was enough to satisfy the tests? (This is definitely one reason to like a testing approach that can be more specific about what kind of error it expects....)

cpovirk: > Perhaps you just need to do [this TODO](https://github.com/jspecify/jspecify-reference-checker/blob/8eb4ab8280e373c6ec97914a0f1fcf1a62a9e32f/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java#L146)?

Sorry, I was looking at `main-eisop` there and not this PR.... Here's where this PR does it. It looks right to me at first blush:

https://github.com/jspecify/jspecify-reference-checker/blob/41a4ce499c3b9f7440cca68f57934958b1fdf599/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java#L154

cpovirk:
> What is strange about this is that there is no `Set.of` in that file: https://github.com/jspecify/jspecify-reference-checker/blob/main-eisop/src/main/java/com/google/jspecify/nullness/NullSpecVisitor.java

I don't know where it's picking the commit from, but the logs show that it's picking 9cfc77d1a477371bcc743b7a3604cd8dde16c2a6, which does at least have the `Set.of` call:

https://github.com/jspecify/jspecify-reference-checker/blob/9cfc77d1a477371bcc743b7a3604cd8dde16c2a6/src/main/java/com/google/jspecify/nullness/NullSpecVisitor.java#L592

cpovirk:
Is this the PR that you were seeing the surprise CI behavior for? (Sorry, I should really remember that....)

I just created 3 local clones:
- a clone of git@github.com:jspecify/jspecify-reference-checker.git at the main head  (75aa1de2a8f88f8c23347d86465844df9887874c)
- a clone of git@github.com:jspecify/jspecify-reference-checker.git at the main-eisop head (8eb4ab8280e373c6ec97914a0f1fcf1a62a9e32f)
- a clone of git@github.com:jspecify/jspecify-reference-checker.git at this PR's head (e073d350613439c06483adec471eff1b92554dd4)

Alongside each, I ran `git clone git@github.com:jspecify/jspecify.git -b samples-google-prototype`. Then, in each, I ran `./gradlew jspecifySamplesTest --include-build ../jspecify`.

- The main run succeeds.
- The main-eisop run (after a detour into https://github.com/jspecify/jspecify-reference-checker/issues/167) produces various failures.
- The PR run produces various failures.

I think that's all expected.

Here are the details in test results:

```patch
Lenient samples
- 524 unexpected diagnostics were found
+ 486 unexpected diagnostics were found

Lenient implementwithnullabletypeargument
+ 1 unexpected diagnostic was found

Lenient wildcardswithdefault
- 3 unexpected diagnostics were found

Strict samples
- 1367 unexpected diagnostics were found
+ 1495 unexpected diagnostics were found

Strict implementwithnullabletypeargument
+ 1 unexpected diagnostic was found

Strict wildcardswithdefault
- 6 unexpected diagnostics were found
+ 4 unexpected diagnostics were found
```

I notice that this differs from the CI results, apparently because the CI for `main-eisop` checks out a different branch of the `jspecify` repo:

https://github.com/jspecify/jspecify-reference-checker/blob/8eb4ab8280e373c6ec97914a0f1fcf1a62a9e32f/.github/workflows/build.yml#L36-L37

Does that all seem reasonable to you? (Again, I might not even be asking about the right PR/branch....)

wmdietl:
Thanks for looking into this @cpovirk !
Yes, this is the branch we had been discussing earlier.

For `main-eisop` and this PR you need to use the `samples-google-prototype-eisop` branch of jspecify, because the error marker format changed.

I'll try to reproduce your numbers. Both locally and on CI, for the lenient test, I see:

````
tests.NullSpecTest$Lenient > run[/home/runner/work/jspecify-reference-checker/jspecify-reference-checker/jspecify-reference-checker/build/conformanceTests/samples] FAILED
    java.lang.AssertionError: 816 out of 1227 expected diagnostics were found.
    3 unexpected diagnostics were found:
...
    411 expected diagnostics were not found:
````

Your numbers look very different, so I'm not sure what's going on.

**Comment by cpovirk** [comment]
Ah, thanks, I should have tried `samples-google-prototype-eisop`. With that, I'm able to reproduce the 816/3/411 numbers for the lenient `samples` test.

With this PR, I:
- added `systemProperties += ['emit.test.debug': 'true']` to the `tasks.withType(Test).configureEach` block in `build.gradle` (not sure if this was necessary)
- ran the tests again
- pulled out the first `javac -processor` line from the output (the one for `samples` in lenient mode
- ran that directly

It ends like this:

```
found   : Lib<? extends Foo?>
required: Lib<? extends Foo>
ContainmentSuper.java:38:46: compiler.err.proc.messager: (type.argument.type.incompatible) $$ 4 $$ A $$ Check $$ Lib<? super Foo> $$ Lib<? super Foo?> $$ ( 1446, 1462 ) $$ incompatible type argument for type parameter A of Check.
found   : Lib<? super Foo>
required: Lib<? super Foo?>
ContainmentSuperVsExtends.java:24:41: compiler.err.proc.messager: (type.argument.type.incompatible) $$ 4 $$ A $$ Check $$ Lib<? super Foo> $$ Lib<? extends Object> $$ ( 885, 901 ) $$ incompatible type argument for type parameter A of Check.
found   : Lib<? super Foo>
required: Lib<? extends Object>
ContainmentSuperVsExtends.java:31:26: compiler.err.proc.messager: (type.argument.type.incompatible) $$ 4 $$ A $$ Check $$ Lib<? super Foo> $$ Lib<? extends Object> $$ ( 1136, 1152 ) $$ incompatible type argument for type parameter A of Check.
found   : Lib<? super Foo>
required: Lib<? extends Object>
ContainmentSuperVsExtendsSameType.java:23:41: compiler.err.not.within.bounds: ContainmentSuperVsExtendsSameType.Lib<? super java.lang.Number>, A
133 errors
4 warnings
- compiler.note.proc.messager:
=====
It has many rough edges.
=====
```

The files are being processed in alphabetical order, so it seems that something goes wrong before we even get to `ContravariantReturns` (which would be the very next file)—and of course many other files.

I wondered if I'd get more information if I went back to the normal, user-facing output, so I removed `-Adetailedmsgtext` (and `-XDrawDiagnostics`, which I think was irrelevant, oops). The output still ended mysteriously after `ContainmentSuperVsExtendsSameType.java`.

Hmm, wait a minute: That's a _javac_ error:

```
/tmp/tmp.Mrxwdt3AcN/main-eisop/jspecify-reference-checker/build/conformanceTests/samples/ContainmentSuperVsExtendsSameType.java:23: error: type argument Lib<? super Number> is not within bounds of type-variable A
    new Check<Lib<? extends Number>, Lib<? super Number>>();
                                        ^
  where A,F are type-variables:
    A extends F declared in class Check
    F extends @org.jspecify.annotations.Nullable Object declared in class Check
1 error
```

So I suspect that https://github.com/jspecify/jspecify/pull/485 broke this.

I guess that the runner used by `jspecifySamplesTest` doesn't abort on javac errors.

**Comment by wmdietl** [comment]
Thanks for tracking this down, @cpovirk !
I'll look into why javac errors aren't surfaced properly and into fixing the problem introduced in https://github.com/jspecify/jspecify/pull/485.

**Comment by wmdietl** [comment]
I opened https://github.com/jspecify/jspecify/pull/498 to fix the compilation error. While doing that I ran into https://github.com/google/error-prone/issues/4343. I ended up disabling error-prone on samples anyways, b/c there are too many warnings for the hand-crafted samples.

With that fixed, the numbers in this PR look much better. I'll still try to first track down why javac errors don't show up in test output - they do show up in stock EISOP.

**Review by wmdietl — COMMENTED** [review]
There is some mismatch in running the conformance tests between my local setup and CI... not sure what's going on.
Once that is done, I suggest we finally merge this. Conformance test scores improve and  #159, #161, and #163 are fixed.
Some of the sample tests produce slightly worse results, but I rather look at those separately.
@cpovirk Could you have a look at this PR and let me know what you think?

**Comment by wmdietl** [comment]
The mismatch is

````
        -PASS: implementWithNullableTypeArgument/implementwithnullabletypeargument/User.java: no unexpected facts
        +FAIL: implementWithNullableTypeArgument/implementwithnullabletypeargument/User.java: no unexpected facts
````

I'm trying to find what is different in CI.

File: src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java [hunk 5]
```diff
@@ -138,15 +144,39 @@ final class NullSpecAnnotatedTypeFactory
...
+        TypeUseLocation.TYPE_VARIABLE_USE,
         TypeUseLocation.OTHERWISE
       };
 
+  private static final TypeUseLocation[] nullMarkedLocationsMinusNull =
+      new TypeUseLocation[] {
+        TypeUseLocation.CONSTRUCTOR_RESULT,
+        TypeUseLocation.EXCEPTION_PARAMETER,
+        TypeUseLocation.IMPLICIT_LOWER_BOUND,
+        TypeUseLocation.RECEIVER,
```
cpovirk, line 164: For...

```
        TypeUseLocation.EXCEPTION_PARAMETER,
        TypeUseLocation.IMPLICIT_LOWER_BOUND,
        TypeUseLocation.RECEIVER,
```

...would it make sense to include those in `defaultLocationsMinusNull`, as well?

src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
```diff
@@ -138,15 +144,39 @@ final class NullSpecAnnotatedTypeFactory
...
+        TypeUseLocation.OTHERWISE
+      };
+  private static final TypeUseLocation[] nullMarkedLocationsUnionNull =
+      new TypeUseLocation[] {
+        TypeUseLocation.LOCAL_VARIABLE,
+        TypeUseLocation.RESOURCE_VARIABLE,
+        TypeUseLocation.IMPLICIT_WILDCARD_UPPER_BOUND_NO_SUPER,
+        TypeUseLocation.IMPLICIT_WILDCARD_UPPER_BOUND_SUPER,
```
cpovirk, line 172: Similarly, I think we could include this _one_ in `defaultLocationsUnionNull`.

src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
```diff
@@ -861,8 +894,8 @@ protected AnnotatedTypeMirror substituteTypeVariable(
         substitute.replaceAnnotation(minusNull);
       } else if (argument.hasAnnotation(unionNull) || use.hasAnnotation(unionNull)) {
         substitute.replaceAnnotation(unionNull);
-      } else if (argument.hasAnnotation(nullnessOperatorUnspecified)
-          || use.hasAnnotation(nullnessOperatorUnspecified)) {
+      } else if (argument.hasEffectiveAnnotation(nullnessOperatorUnspecified)
```
cpovirk, line 921: I'm going to just not think about this part for now, given the chance that our substitution behavior will change in one or more ways (https://github.com/jspecify/jspecify/issues/248#issuecomment-2008031089, https://github.com/jspecify/jspecify/issues/91#issuecomment-2008104170) and given [my lack of deep understanding of the Checker Framework model](https://docs.google.com/document/d/1p3S3LR8FPJ7YGVjeZVzTcLOCIeznBl4dx5fAfdf-oR4/edit?resourcekey=0-wcbLq8RMeA_ujzNG5aif_Q) :)

cpovirk (approved):

cpovirk:
I kicked off an internal build with the current code in this PR (752ad0978019cdfdbe6675ffa807a1486e55858e). I'll have to sort through the result in more detail on another day, but the main thing that jumps out to me at the moment is a lot of messages along the lines of `found: Foo<T> | required: Foo<T>`. It's somewhat like https://github.com/jspecify/jspecify-reference-checker/issues/164 but not resricted to `? super`:

```
java/com/google/common/collect/HashBiMap.java:707: error: [override.param.invalid] Incompatible parameter type for entry.
        Entry<V, K> output(BiEntry<K, V> entry) {
                                         ^
  found   : BiEntry<K, V>
  required: BiEntry<K, V>
  Consequence: method in 
    Entry<V, K> output(BiEntry<K, V>)
  cannot override method in Itr<Entry<V, K>>
    Entry<V, K> output(BiEntry<K, V>)
```

src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
wmdietl: I must be missing something... that suggestion does make sense, but that is already what `defaultLocationsMinusNull` further up in this file contains.
What change are you suggesting? Is there some documentation I can add to make these easier to read?

cpovirk: Sorry, I'm not sure what I was looking at. It looks like GitHub's review UI is collapsing those lines by default, so I probably thought that `CONSTRUCTOR_RESULT` was the only entry in that list. Never mind.

I _think_ my similar comment below may still actually be correct, but let me know if not.

wmdietl: `IMPLICIT_WILDCARD_UPPER_BOUND_SUPER` is also already in `defaultLocationsUnionNull`.
It's sometimes easiest to click on the `...` on the top right of a file and click `View file` to look at the whole file.

cpovirk: Sigh, sorry again.

src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
cpovirk: (oops—discussed above)

wmdietl:
> I kicked off an internal build with the current code in this PR ([752ad09](https://github.com/jspecify/jspecify-reference-checker/commit/752ad0978019cdfdbe6675ffa807a1486e55858e)). I'll have to sort through the result in more detail on another day, but the main thing that jumps out to me at the moment is a lot of messages along the lines of `found: Foo<T> | required: Foo<T>`. It's somewhat like #164 but not resricted to `? super`:

Thanks for re-running your tests!
I think what is happening is that the `@NullMarked` is ignored when code comes from an `Element`, that is, a different compilation unit or bytecode.
We treat `@NullMarked` and `@NullUnmarked` as changes to defaults. Defaults are however only applied when we check source code. In stock CF, we write the defaulted types back into the `Element`, which then allows us to directly see the annotations in bytecode.
The reference checker intentionally does not store defaults back into bytecode. I think we need to change how we load bytecode and similarly look for `@Null(Un)Marked` on enclosing elements.
Similarly, when looking at stub files, stock CF only looks at `@AnnotatedFor`. I'm not sure the defaults set by `@Null(Un)Marked` have any effect in stubs and the annotated JDK.
Does either of these sound like you had work-arounds for them before?
I don't see how this PR would have made that behavior different, so I think we can still merge this.

Because the `@NullMarked` is ignored, the upper bounds for these type variables are `UnknownNullness` and therefore the override check fails. But fixing the override check, ala #164, is probably not enough and we need to fix the above handling of defaults.
(The error message is pretty cryptic, as it's the same type twice. The reference checker intentionally doesn't show bounds of type variables. I believe that in the bound we would see unknown nullness. Then we would still see the same type twice, but at least if one remembers that subtyping for unknown nullness is not reflexive, it might make sense.)

However, I'm still trying to figure out why the conformance test is different on CI than on my local machine...

cpovirk (approved):
Good info, thanks.

I'm still somewhat confused about the errors: If the problem were merely that we're seeing accidentally unspecified types, then I'd expect _not_ to see errors, since we're running in lenient mode inside Google. But it's possible that we're seeing accidentally unspecified types _and_ that we're not handling subtyping (and/or "same-type" checking?) correctly in the presence of unspecified types: As you say, we already have seen a similar-looking problem in #164.

As for workarounds: We certainly do have [our own separate defaulting system](https://github.com/jspecify/jspecify-reference-checker/blob/75aa1de2a8f88f8c23347d86465844df9887874c/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java#L814). There seem to have been multiple reasons that led me down that path. I don't specifically remember being aware of the problem that you mentioned, but it's been a long time, so maybe I've just forgotten. (I could dig into the commit history, but it's probably academic.)

The rule of "look for defaults only in source code" probably also explains why I had to hack up the stub-file code so badly in order to recognize `@NullMarked` in the JDK stubs. Fortunately, I think we can get away without worrying about stub files for now. But it is good to have a deeper understanding of why that was hard to get working.

I'm fine with merging to lock in the progress so far.

cpovirk:
Some notes:

- `Itr` comes from the same compilation unit as the subclass: Both are in [`HashBiMap.java`](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/HashBiMap.java). So I don't think that a difference in defaulting between source files and class files explains it.
- I'm also still unclear on "Defaults are however only applied when we check source code." I traced to https://github.com/eisop/checker-framework/blob/ddb3e090073311d88a0a2300ab5b4da8bed124c5/framework/src/main/java/org/checkerframework/framework/util/defaults/QualifierDefaults.java#L425-L461 and then to https://github.com/eisop/checker-framework/blob/ddb3e090073311d88a0a2300ab5b4da8bed124c5/framework/src/main/java/org/checkerframework/framework/util/defaults/QualifierDefaults.java#L745 (noting that `getDeclAnnotation` says it supports aliases), and that seems to support the idea that we could read defaults from class files (not source-retention `@DefaultQualifier` itself but at least class/runtime-retention aliases like `@NullMarked`). And when I added println debugging to `defaultsAtDirect(Element elt) `, I saw it called for types like `java.lang.Object`. (That doesn't mean that it _works_ :) I didn't investigate further. But the checker _does_ seem to be _trying_.) If I'm right about my first bullet, then this is academic for now, but it might matter for later.
- Here's a minimized version:

```java

@NullMarked
class HashBiMapAlike<E extends @Nullable Object> {
  abstract class Super {
    abstract void accept(Entry<E> entry);
  }

  abstract class Sub extends Super {
    @Override
    abstract void accept(Entry<E> entry);
  }
}

interface Entry<E extends @Nullable Object> {}
```

```
$ ./demo HashBiMapAlike.java 
HashBiMapAlike.java:12: error: [override.param.invalid] Incompatible parameter type for entry.
    abstract void accept(Entry<E> entry);
                                  ^
  found   : Entry<E>
  required: Entry<E>
  Consequence: method in Sub
    void accept(Entry<E>)
  cannot override method in Super
    void accept(Entry<E>)
1 error
```
- I confirmed that the above test passes at `main-eisop` (using my instructions from https://github.com/jspecify/jspecify-reference-checker/pull/165#issuecomment-2013277831), but that still leaves the question of what sort of changes to defaulting or `@ParametricNull` might be the cause.

cpovirk:
tl;dr I think the conformance-test runner needs to sort the input files. Not that that _should_ matter, but I've seen it matter for my original `samples` integration, which I have set up to sort files.

Details:

I mentioned this part on an internal doc, but I'll say it here, too: The tests pass for me locally, as well. (That includes if I use JDK17 and if I use the `./gradlew build conformanceTests demoTest --include-build ../jspecify` command line from CI.)

I wonder if we're seeing something similar to https://github.com/jspecify/jspecify-reference-checker/pull/165#issuecomment-2015364311, in which an error from javac itself stops us from checking many files. And maybe our different systems have very slight difference in how javac behaves, either because of different versions or because of environmental perturbations, perhaps the most likely of which would be filesystem iteration order?? Even if we're not seeing a javac error, I have seen cases (unfortunately) in which the order of files matters.

To test that, I just tried a run in which the conformance tests use `ImmutableSortedSet` instead of `ImmutableList` for their input files. And... that produced the failure locally!

```
    diff (-expected +actual):
        @@ -1,4 +1,4 @@
        -# 73 pass; 500 fail; 573 total; 12.7% score
        +# 72 pass; 501 fail; 573 total; 12.6% score
         FAIL: AnnotatedInnerOfNonParameterized.java: no unexpected facts
         FAIL: AnnotatedInnerOfParameterized.java: no unexpected facts
         FAIL: AnnotatedReceiver.java: no unexpected facts
        @@ -532,7 +532,7 @@
         FAIL: ignoreAnnotations/ignoreannotations/IgnoreAnnotations.java: no unexpected facts
         PASS: implementWithNullableTypeArgument/implementwithnullabletypeargument/MyFunction.java: no unexpected facts
         PASS: implementWithNullableTypeArgument/implementwithnullabletypeargument/SuperFunction.java: no unexpected facts
        -PASS: implementWithNullableTypeArgument/implementwithnullabletypeargument/User.java: no unexpected facts
        +FAIL: implementWithNullableTypeArgument/implementwithnullabletypeargument/User.java: no unexpected facts
         PASS: memberSelectNonExpression/memberselectnonexpression/MemberSelectNonExpressions.java: no unexpected facts
         PASS: nonPlatformTypeParameter/nonplatformtypeparameter/NonPlatformTypeParameter.java: no unexpected facts
         PASS: nullnessUnspecifiedTypeParameter/nullnessunspecifiedtypeparameter/NullnessUnspecifiedTypeParameter.java:35:jspecify_nullness_mismatch
```

The extra error, incidentally, appears to be this:

```
OOPS: implementWithNullableTypeArgument/implementwithnullabletypeargument/User.java:25:test:cannot-convert:SuperFunction!<Object!, Object!> to Object apply(Object)!
```

```
2024-03-28T14:07:45.7406117Z       /home/runner/work/jspecify-reference-checker/jspecify-reference-checker/jspecify-reference-checker/build/conformanceTests/samples/implementWithNullableTypeArgument/implementwithnullabletypeargument/User.java:25: (override.return.invalid) Incompatible return type.
2024-03-28T14:07:45.7409196Z         @Nullable Object apply(@Nullable Object f);
2024-03-28T14:07:45.7409908Z                   ^
2024-03-28T14:07:45.7410397Z       found   : Object?
2024-03-28T14:07:45.7410929Z       required: Object
```

That looks like a substitution problem. (Sorry, I can't seem to remember whether your existing suspicion had been a substitution problem or not.)

Since there are multiple files involved in that sample, the ordering hypothesis is relatively plausible. And sure enough:

```
$ ./demo ../jspecify/samples/implementWithNullableTypeArgument/implementwithnullabletypeargument/SuperFunction.java ../jspecify/samples/implementWithNullableTypeArgument/implementwithnullabletypeargument/User.java ../jspecify/samples/implementWithNullableTypeArgument/implementwithnullabletypeargument/MyFunction.java 

(no errors)

$ ./demo ../jspecify/samples/implementWithNullableTypeArgument/implementwithnullabletypeargument/MyFunction.java ../jspecify/samples/implementWithNullableTypeArgument/implementwithnullabletypeargument/SuperFunction.java ../jspecify/samples/implementWithNullableTypeArgument/implementwithnullabletypeargument/User.java 
../jspecify/samples/implementWithNullableTypeArgument/implementwithnullabletypeargument/User.java:25: error: [override.return.invalid] Incompatible return type.
    @Nullable Object apply(@Nullable Object f);
              ^
  found   : Object?
  required: Object
  Consequence: method in NullableObjectFunction
    Object? apply(Object?)
  cannot override method in SuperFunction<Object, Object>
    Object apply(Object)
1 error
```

tests/minimal/Issue159.java
```diff
@@ -0,0 +1,26 @@
+// Copyright 2024 The JSpecify Authors
```
wmdietl: @cpovirk @netdpb  Let's re-start the discussion from https://github.com/jspecify/jspecify-reference-checker/pull/165#discussion_r1503315408

I've added tests for the three fixed issues and the override test from one of the last comments. I could either remove them before we merge this or maybe move them from `minimal` to `issues` or some other sub-directory?

We can then merge this and I'll work on #164 in a new PR.

wmdietl: We discussed this today and settled on moving the tests. Done in https://github.com/jspecify/jspecify-reference-checker/pull/165/commits/2588340d48ce1e94fc906d6f060c6fe069f32444 
A future cleanup can move some of them to conformance tests and keep others.

PR #168, opened 2024-04-02
Sort input files in the test runner.

cpovirk:
While the order of the input files _shouldn't_ affect behavior, it
sometimes does. Let's use a consistent order so that we at least don't
see behavior differences between local and CI runs:

https://github.com/jspecify/jspecify-reference-checker/pull/165#issuecomment-2032659286

### DIFF
--- a/conformance-test-framework/src/main/java/org/jspecify/conformance/ConformanceTestRunner.java
+++ b/conformance-test-framework/src/main/java/org/jspecify/conformance/ConformanceTestRunner.java
@@ -27,6 +27,7 @@ import static java.util.stream.Collectors.toList;
 
 import com.google.common.base.Ascii;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSortedSet;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.file.Files;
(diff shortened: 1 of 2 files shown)


conformance-test-framework/src/main/java/org/jspecify/conformance/ConformanceTestRunner.java
```diff
@@ -77,7 +78,7 @@ public ConformanceTestReport runTests(Path testDirectory, ImmutableList<Path> te
                     ? groups.flatMap(files -> partition(files, 1).stream())
                     : groups;
               })
-          .map(ImmutableList::copyOf)
+          .map(ImmutableSortedSet::copyOf)
```
netdpb, line 82: Maybe change `javaFileGroups(Path)` to return `Stream<ImmutableSortedSet<Path>>` and `javaFilesInDirectory(Path)` to return `ImmutableSortedSet<Path>` instead?

You'd have to do something with the partitioning in line 78 like `files -> files.stream().map(ImmutableSortedSet::of)`.

netdpb (approved):

conformance-test-framework/src/main/java/org/jspecify/conformance/ConformanceTestRunner.java
cpovirk: Done.

PR #171, opened 2024-04-12
Ensure override compatibility with type variables

### LINKED ISSUE #164: main-eisop: `x(Foo<? super T>)` cannot override identical signature

wmdietl:
I want to look whether there is a better fix for this, maybe by moving this check into the override checker.
Instead, maybe something is wrong with the logic in the `return` statement that follows this change.
Let me know if this looks familiar.

Fixes #164.

### DIFF
--- a/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
+++ b/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
@@ -667,7 +667,10 @@ final class NullSpecAnnotatedTypeFactory
         // Work around wonkyness of CF override checks.
         AnnotatedTypeVariable subTV = (AnnotatedTypeVariable) subtype;
         if (isCapturedTypeVariable(subTV.getUnderlyingType())) {
-          subTV = (AnnotatedTypeVariable) subTV.getUpperBound();
+          AnnotatedTypeMirror subTVBnd = subTV.getUpperBound();
+          if (subTVBnd instanceof AnnotatedTypeVariable) {
+            subTV = (AnnotatedTypeVariable) subTVBnd;
+          }
         }
         AnnotatedTypeVariable superTV = (AnnotatedTypeVariable) supertype;
         if (areCorrespondingTypeVariables(elements, subTV, superTV)) {
--- a/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
+++ b/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
@@ -664,7 +664,11 @@ final class NullSpecAnnotatedTypeFactory
         return true;
       }
       if (subtype.getKind() == TYPEVAR && supertype.getKind() == TYPEVAR) {
+        // Work around wonkyness of CF override checks.
         AnnotatedTypeVariable subTV = (AnnotatedTypeVariable) subtype;
+        if (isCapturedTypeVariable(subTV.getUnderlyingType())) {
+          subTV = (AnnotatedTypeVariable) subTV.getUpperBound();
+        }
         AnnotatedTypeVariable superTV = (AnnotatedTypeVariable) supertype;
         if (areCorrespondingTypeVariables(elements, subTV, superTV)) {
           return isSubtype(subTV.getUpperBound(), superTV.getUpperBound())
--- a/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
+++ b/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
@@ -31,6 +31,7 @@ import static javax.lang.model.type.TypeKind.ARRAY;
 import static javax.lang.model.type.TypeKind.DECLARED;
 import static javax.lang.model.type.TypeKind.TYPEVAR;
 import static javax.lang.model.type.TypeKind.WILDCARD;
+import static org.checkerframework.framework.util.AnnotatedTypes.areCorrespondingTypeVariables;
 import static org.checkerframework.framework.util.AnnotatedTypes.asSuper;
 import static org.checkerframework.javacutil.AnnotationUtils.areSame;
 import static org.checkerframework.javacutil.TreePathUtil.enclosingClass;
@@ -639,6 +640,14 @@ final class NullSpecAnnotatedTypeFactory
           && !supertype.hasAnnotation(minusNull)
           && isNullnessSubtype(subtype, ((AnnotatedTypeVariable) supertype).getLowerBound())) {
         return true;
+      }
+      if (subtype.getKind() == TYPEVAR && supertype.getKind() == TYPEVAR) {
+        AnnotatedTypeVariable subTV = (AnnotatedTypeVariable) subtype;
+        AnnotatedTypeVariable superTV = (AnnotatedTypeVariable) supertype;
+        if (areCorrespondingTypeVariables(elements, subTV, superTV)) {
+          return isSubtype(subTV.getUpperBound(), superTV.getUpperBound())
+              && isSubtype(superTV.getLowerBound(), subTV.getLowerBound());
+        }
       }
       return isNullInclusiveUnderEveryParameterization(supertype)
           || isNullExclusiveUnderEveryParameterization(subtype)
--- a/tests/ConformanceTestOnSamples-report.txt
+++ b/tests/ConformanceTestOnSamples-report.txt
@@ -1,4 +1,4 @@
-# 73 pass; 500 fail; 573 total; 12.7% score
+# 74 pass; 499 fail; 573 total; 12.9% score
 FAIL: AnnotatedInnerOfNonParameterized.java: no unexpected facts
 FAIL: AnnotatedInnerOfParameterized.java: no unexpected facts
 FAIL: AnnotatedReceiver.java: no unexpected facts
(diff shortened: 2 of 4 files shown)


cpovirk (approved):
I will organize some thoughts, but I am all for merging even a hacky fix.

cpovirk:
Again, LGTM.

I will test this in the Google codebase and report back, but I don't think that even that is a blocker for merging. (If I discover that things somehow break as a result, we can always revert the commit :))

My next thought is that it does seem that this new case "shouldn't be necessary": `nullnessEstablishingPathExists` should find the "path" from `T` to `T`.

Ah, I wonder if somehow those are two different `T` types, as if the step of "adapting the formal parameter types of N to the type parameters of M" from [JLS 8.4.2](https://docs.oracle.com/javase/specs/jls/se22/html/jls-8.html#jls-8.4.2) is not being applied? That would be my guess: I seem to recall that the Checker Framework will let you override a supermethod `<T extends Object> void foo()` with a submethod `<T extends @Nullable Object> void foo()`, similar to how it allows parameter contravariance. Presumably that would require holding two different concepts of `T` in mind simultaneously when checking that any formal parameters have covariant types. (This is fresh on my mind because I was [spreading FUD about it yesterday](https://github.com/jspecify/jspecify/issues/49#issuecomment-2050541668) :)) So the Checker Framework may be checking overrides in a fundamentally different want than the JLS does... but I guess still delegating to `isSubtype` as part of that? I think that may line up well with my understanding of your fix. Or I could have any number of things totally wrong :)

Anyway, I've now talked myself into at least sort of having a theory about what's going on. And if that theory is right, your fix is probably the best we can do on the jspecify-reference-checker side. If so, it's still _possible_ that something more can be done on the checker-framework side. But I wouldn't worry about it: There's still a decent chance that we end up redoing some of this code as a result of changes in design decisions and/or to make it fit more naturally into the Checker Framework model (instead of having this extensive `isSubtype` logic in jspecify-reference-checker at all).

cpovirk:
It looks like I may be seeing _fewer_ "incompatible parameter type" errors, albeit still not zero. It's a bit hard to tell how much of the reduction is because of #159 crashes that hide other problems, but clearly at least some of the reduction is real.

An example of a line on which I'm still seeing a problem is this one:

https://github.com/google/guava/blob/858caf425cdf3183776b61b437ba7a4129b6539b/guava/src/com/google/common/collect/MultimapBuilder.java#L472

Both the "found" type and the "required" type are "Multimap<? extends K, ? extends V>."

cpovirk:
The remaining issue might be restricted to type parameters that extend other type parameters (and, similarly, _wildcards_ that extend type parameters). Here's an example:

```java

@NullMarked
class Issue164Continued {
  interface Super<T extends @Nullable Object> {
    <U extends T> void foo(Lib<? extends U> lib);
  }

  interface Sub<T extends @Nullable Object> extends Super<T> {
    <U extends T> void foo(Lib<? extends U> lib);
  }

  interface Lib<T extends @Nullable Object> {}
}
```

```
$ ./demo Issue164Continued.java
Issue164Continued.java:11: error: [override.param.invalid] Incompatible parameter type for lib.
    <U extends T> void foo(Lib<? extends U> lib);
                                            ^
  found   : Lib<? extends U>
  required: Lib<? extends U>
  Consequence: method in Sub<T>
    void foo(Lib<? extends U>)
  cannot override method in Super<T>
    void foo(Lib<? extends U>)
1 error
```

It's a rare enough problem that I'd feel OK with adding suppressions for it. (I could believe that a proper fix would involve _more_ `areCorrespondingTypeVariables`+`isSubtype` checks recursively on the bounds when the bounds are themselves type parameters.)

cpovirk:
(Well, I guess we'll see how common it _really_ is after the crashes are gone. Maybe it's worse than I think.)

cpovirk:
At a higher level:

- In addition to helping with various cases, this PR doesn't make anything worse (as far as I can tell).
- If we merge it, and if we then find a fix for [#159](https://github.com/jspecify/jspecify-reference-checker/issues/159), and if nothing new breaks at that point, and if the crashes from [#159](https://github.com/jspecify/jspecify-reference-checker/issues/159) aren't hiding anything big, then I think we'd be ready to switch Google over, as the problems I reported [above](https://github.com/jspecify/jspecify-reference-checker/pull/171#issuecomment-2052343848) and in [#172](https://github.com/jspecify/jspecify-reference-checker/issues/172) appear to be minor.
- To get a preview of all that, I applied the hacky checker-framework patch from [my comment on #159](https://github.com/jspecify/jspecify-reference-checker/issues/159#issue-2125283123) to silence the crashes. I didn't notice any new kinds of errors, and the level of errors does appear tolerable. (And I encountered at least [a couple errors](https://github.com/google/guava/pull/7156) that were legitimate (if probably harmless), too!)

wmdietl:
@cpovirk Thanks for the additional test!
The override checks seem a bit odd, but let's not try to fight them now. Adding an additional check for the subtype type variable being captured, in https://github.com/jspecify/jspecify-reference-checker/pull/171/commits/9421ef85843ad589c578c721aa35731fd6b90cc2, fixed the additional test and doesn't change the conformance results.

I like using different type variable names for different type variable declarations. That makes it easier to see in messages that it is looking at different types. The `areCorrespondingTypeVariables` check makes sure the different TVs from the two methods correspond.

If this is producing okay results, I would propose merging and cleaning up the CF interaction and wonkiness separately.

src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
```diff
@@ -664,7 +664,11 @@ && isNullnessSubtype(subtype, ((AnnotatedTypeVariable) supertype).getLowerBound(
         return true;
       }
       if (subtype.getKind() == TYPEVAR && supertype.getKind() == TYPEVAR) {
+        // Work around wonkyness of CF override checks.
         AnnotatedTypeVariable subTV = (AnnotatedTypeVariable) subtype;
+        if (isCapturedTypeVariable(subTV.getUnderlyingType())) {
+          subTV = (AnnotatedTypeVariable) subTV.getUpperBound();
```
cpovirk: It looks like we'd need to check that assumption:

> Exception: java.lang.ClassCastException: class org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType cannot be cast to class org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedTypeVariable

That error is probably happening at https://github.com/google/auto/blob/68d99dc75b8379ca14e56c8c4f10493b1b0cb38d/common/src/main/java/com/google/auto/common/SuperficialValidation.java#L237.

I think we're fine just not worrying about trying to handle the captured-type-variable case at all for now. But if you want to look into it more, I can get you more information tomorrow. Let me know.

cpovirk (approved):

src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
wmdietl: Ah! I added a check to ensure the cast is safe.

cpovirk: I've confirmed (a) that it eliminates all the crashes I was seeing and (b) that the overall change does address the `? extends E` errors I was seeing.

PR #173, opened 2024-04-14
Test fixes

wmdietl:
CI fails because of #159.
This PR fixes the test setup to actually highlight that failure.

### DIFF
--- a/src/test/java/tests/DetailMessage.java
+++ b/src/test/java/tests/DetailMessage.java
@@ -55,13 +55,12 @@ final class DetailMessage extends TestDiagnostic {
   final String readableMessage;
 
   /**
-   * Returns an object parsed from a diagnostic message, or {@code null} if the message doesn't
-   * match the expected format.
+   * Returns an object parsed from a diagnostic message.
    *
    * @param rootDirectory if not null, a root directory prefix to remove from the file part of the
    *     message
    */
-  static @Nullable DetailMessage parse(TestDiagnostic input, @Nullable Path rootDirectory) {
+  static DetailMessage parse(TestDiagnostic input, @Nullable Path rootDirectory) {
     Matcher detailsMatcher = DETAIL_MESSAGE_PATTERN.matcher(input.getMessage());
     if (!detailsMatcher.matches()) {
       // Return a message with no key or parts.
@@ -142,7 +141,8 @@ final class DetailMessage extends TestDiagnostic {
       return false;
     }
     DetailMessage that = (DetailMessage) o;
-    return super.equals(that)
+    return lineNumber == that.lineNumber
+        && file.equals(that.file)
         && messageKey.equals(that.messageKey)
         && messageArguments.equals(that.messageArguments)
         && Objects.equals(offsetStart, that.offsetStart)
@@ -153,7 +153,7 @@ final class DetailMessage extends TestDiagnostic {
   @Override
   public int hashCode() {
     return Objects.hash(
-        super.hashCode(), messageKey, messageArguments, offsetStart, offsetEnd, readableMessage);
+        file, lineNumber, messageKey, messageArguments, offsetStart, offsetEnd, readableMessage);
   }
 
   @Override
--- a/src/test/java/tests/DetailMessage.java
+++ b/src/test/java/tests/DetailMessage.java
@@ -62,8 +62,6 @@ final class DetailMessage extends TestDiagnostic {
    *     message
    */
   static @Nullable DetailMessage parse(TestDiagnostic input, @Nullable Path rootDirectory) {
-    if (input.getFile().toString().contains("CurrentThreadThreadGroup"))
-      System.out.println("looking at: " + input);
     Matcher detailsMatcher = DETAIL_MESSAGE_PATTERN.matcher(input.getMessage());
     if (!detailsMatcher.matches()) {
       // Return a message with no key or parts.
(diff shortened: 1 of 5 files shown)


wmdietl:
@netdpb The conformance tests suddenly accept `DetailMessages` with compilation errors. I verified that the compilation errors are reported in `analyze`, with kind error. I haven't tracked down where they are filtered out.
I don't think https://github.com/eisop/checker-framework/pull/739 is the problem, but that's a possibility.

`./gradlew test` fails in this PR, so it might be easier to run in #174.

There, you'll e.g. see that `FAIL: AnnotatedReceiver.java: no unexpected facts` changed into `PASS: AnnotatedReceiver.java: no unexpected facts` even though one sees the compilation failure when checking with `demo`.

cpovirk:
It looks like the compile error may be getting reported under *full* path but that the conformance-test runner may be looking for errors only in the *relative* paths:

```patch
@@ -73,6 +75,10 @@ public final class ConformanceTestReport {
     report.format(
         "# %,d pass; %,d fail; %,d total; %.1f%% score%n",
         passes, fails, total, 100.0 * passes / total);
+    Set<Path> unknownFilesWithErrors = difference(reportedFactsByFile.keySet(), files);
+    if (!unknownFilesWithErrors.isEmpty()) {
+      throw new RuntimeException("unknown files with errors: " + unknownFilesWithErrors);
+    }
     for (Path file : files) {
       ImmutableListMultimap<Long, ExpectedFact> expectedFactsInFile =
           index(expectedFactsByFile.get(file), Fact::getLineNumber);
```

```
java.lang.RuntimeException: unknown files with errors: [, /tmp/tmp.b0K8g39UiV/jspecify-reference-checker/build/conformanceTests/samples/ContainmentExtends.java, /tmp/tmp.b0K8g39UiV/jspecify-reference-checker/build/conformanceTests/samples/TypeVariableUnionNullToObjectUnionNull.java, ...
```

wmdietl:
Ah! I return the first DetailMessage before relativizing!
I'll see whether that fixes things.

wmdietl:
@cpovirk Thanks for tracking this down! Relativizing earlier fixes the issue.
Should I file against the conformance test suite to make this more robust? Or is this working as intended?

src/test/java/tests/DetailMessage.java
```diff
@@ -71,30 +55,30 @@ final class DetailMessage extends TestDiagnostic {
...
-    Matcher messageMatcher = MESSAGE_PATTERN.matcher(input);
-    if (!messageMatcher.matches()) {
-      return null;
-    }
-
-    Path file = Paths.get(messageMatcher.group("file"));
-    if (rootDirectory != null) {
+  static DetailMessage parse(TestDiagnostic input, @Nullable Path rootDirectory) {
+    Path file = input.getFile();
+    if (rootDirectory != null && !file.toString().equals("")) {
```
netdpb: When would `file` be empty?

netdpb:
> @cpovirk Thanks for tracking this down! Relativizing earlier fixes the issue. Should I file against the conformance test suite to make this more robust? Or is this working as intended?

I think the point of that `rootDirectory` parameter is to relativize. How would you make it more robust?

src/test/java/tests/DetailMessage.java
cpovirk: I can confirm that it was happening, at least :) I didn't investigate further, but I assumed it came from this:

https://github.com/jspecify/jspecify-reference-checker/blob/fae43cfec3b0d9e70cb02ef2d73fa7e05653d4b6/src/main/java/com/google/jspecify/nullness/NullSpecChecker.java#L83-L90

wmdietl: Every test that has an error also produces a note:

````
Note: 
  =====
  It has many rough edges.
  =====
````

That diagnostics has no file path.

**Comment by wmdietl** [comment]
> > @cpovirk Thanks for tracking this down! Relativizing earlier fixes the issue. Should I file against the conformance test suite to make this more robust? Or is this working as intended?
> 
> I think the point of that `rootDirectory` parameter is to relativize. How would you make it more robust?

Output diagnostics even if they are not relative to the `rootDirectory`?
When is it okay for `unknownFilesWithErrors` from Chris's [comment](https://github.com/jspecify/jspecify-reference-checker/pull/173#issuecomment-2059517842) to be non-empty?

**Comment by netdpb** [comment]
> > > @cpovirk Thanks for tracking this down! Relativizing earlier fixes the issue. Should I file against the conformance test suite to make this more robust? Or is this working as intended?
> > 
> > 
> > I think the point of that `rootDirectory` parameter is to relativize. How would you make it more robust?
> 
> Output diagnostics even if they are not relative to the `rootDirectory`? When is it okay for `unknownFilesWithErrors` from Chris's [comment](https://github.com/jspecify/jspecify-reference-checker/pull/173#issuecomment-2059517842) to be non-empty?

It shouldn't ever be. Maybe unexpected errors in unexpected locations (or no location) should cause the runner to throw an exception? I don't want to put those in the test report.

Anyway, feel free to file a separate issue for what to do in that case.

File: src/test/java/tests/DetailMessage.java (hunk 1 above)
Comment (netdpb) [inline src/test/java/tests/DetailMessage.java]:
```suggestion
    if (rootDirectory != null && file.startsWith(rootDirectory)) {
```

I think this is more robust. I think `relativize()` throws if `file` isn't within `rootDirectory`.

netdpb (approved):

wmdietl:
> > > > @cpovirk Thanks for tracking this down! Relativizing earlier fixes the issue. Should I file against the conformance test suite to make this more robust? Or is this working as intended?
> > > 
> > > 
> > > I think the point of that `rootDirectory` parameter is to relativize. How would you make it more robust?
> > 
> > 
> > Output diagnostics even if they are not relative to the `rootDirectory`? When is it okay for `unknownFilesWithErrors` from Chris's [comment](https://github.com/jspecify/jspecify-reference-checker/pull/173#issuecomment-2059517842) to be non-empty?
> 
> It shouldn't ever be. Maybe unexpected errors in unexpected locations (or no location) should cause the runner to throw an exception? I don't want to put those in the test report.
> 
> Anyway, feel free to file a separate issue for what to do in that case.

Filed #176.

src/test/java/tests/DetailMessage.java
```diff
@@ -62,7 +62,7 @@ final class DetailMessage extends TestDiagnostic {
    */
   static DetailMessage parse(TestDiagnostic input, @Nullable Path rootDirectory) {
     Path file = input.getFile();
-    if (rootDirectory != null && !file.toString().equals("")) {
+    if (rootDirectory != null && file.startsWith(rootDirectory)) {
       // Empty file strings cannot be relativized.
```
netdpb: I think you can remove this comment.

PR #181, opened 2024-06-26
Stop recognizing annotations from the old package.

cpovirk:


netdpb:
What about the others, [here](https://github.com/jspecify/jspecify-reference-checker/blob/main/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java#L1829-L1856)?

cpovirk:
Oopsie. Thanks. Done.

src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
```diff
@@ -1983,7 +1975,6 @@ private enum Present {
                 "org.jetbrains.annotations.Nullable",
                 "org.jmlspecs.annotation.Nullable",
                 "org.jspecify.annotations.Nullable",
-                "org.jspecify.nullness.Nullable",
```
netdpb, line 1986: What's up with these entire lists? If this is the reference checker for JSpecify annotations, should it understand anything else at all?

netdpb (approved):

src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
cpovirk: Do you have in mind downsides to recognizing other annotations? The checker already goes beyond the JSpecify spec by handling things like dataflow, and of course we use it on some projects like Guava for our own purposes. I could imagine putting all that logic behind flags (so that the conformance tests could disable it), but I'm not sure what we'd get out of it.

PR #183, opened 2024-07-09
Disable checking of contracts.

cpovirk:
We've historically accomplished this with `-AsuppressWarnings`, but this
CL's solution may be more general. Also, it avoids running some code
entirely, which saves us from having to keep that code updated for the
very new JDKs that we run with. (It might also help performance if we're
lucky, and maybe it moves us a step closer to not needing JavaParser
(though that may be on the way out, anyway).)


cushon (approved):

cpovirk:
Merging despite pre-existing CI failures, for which I've filed https://github.com/jspecify/jspecify-reference-checker/issues/184 and created https://github.com/jspecify/jspecify-reference-checker/pull/185.

Issue #184, opened 2024-07-09
Conformance-test report not automatically regenerated after jspecify change


cpovirk:
The action seems to have [run](https://github.com/jspecify/jspecify/actions/runs/9861656304/job/27230553159) for [#508](https://github.com/jspecify/jspecify/pull/508), but I don't see a resulting PR. And a clean run...

```shell
git clone git@github.com:jspecify/jspecify-reference-checker.git && git clone git@github.com:jspecify/jspecify.git && cd jspecify-reference-checker/ && JAVA_HOME=$HOME/jdk-17.0.2 ./gradlew build conformanceTests demoTest --include-build ../jspecify
```

...fails:

```
tests.ConformanceTest > conformanceTests FAILED
    diff (-expected +actual):
        @@ -1,4 +1,4 @@
        -# 90 pass; 24 fail; 114 total; 78.9% score
        +# 93 pass; 24 fail; 117 total; 79.5% score
         PASS: Basic.java:28:test:expression-type:Object?:nullable
         PASS: Basic.java:28:test:sink-type:Object!:return
         PASS: Basic.java:28:test:cannot-convert:Object? to Object!
        @@ -44,6 +44,7 @@
         PASS: irrelevantannotations/notnullmarked/Other.java:NonNull local variable array:test:irrelevant-annotation:NonNull
         PASS: irrelevantannotations/notnullmarked/Other.java:Nullable exception parameter:test:irrelevant-annotation:Nullable
         PASS: irrelevantannotations/notnullmarked/Other.java:NonNull exception parameter:test:irrelevant-annotation:NonNull
        +PASS: irrelevantannotations/notnullmarked/Other.java:Intrinsically NonNull exception parameter cannot be assigned null:test:cannot-convert:null? to RuntimeException!
         PASS: irrelevantannotations/notnullmarked/Other.java:Nullable try-with-resources:test:irrelevant-annotation:Nullable
         FAIL: irrelevantannotations/notnullmarked/Other.java:NonNull try-with-resources:test:irrelevant-annotation:NonNull
         FAIL: irrelevantannotations/notnullmarked/Other.java:Nullable exception type:test:irrelevant-annotation:Nullable
        @@ -75,6 +76,7 @@
         PASS: irrelevantannotations/nullmarked/Other.java:NonNull local variable array:test:irrelevant-annotation:NonNull
[...]

PR #185, opened 2024-07-09
Update conformance tests for https://github.com/jspecify/jspecify/pull/508

cpovirk:
This hasn't happened automatically, as discussed in
https://github.com/jspecify/jspecify-reference-checker/issues/184.


netdpb (approved):
I noticed that too. Thanks for doing this!

And our score went up!

PR #186, opened 2024-07-10
Use `DetailedTestDiagnostic` from EISOP instead of `DetailMessage`

wmdietl:
This goes together with https://github.com/eisop/checker-framework/pull/758.
Parsing the detailed test diagnostics is also needed for the [EISOP JSpecify conformance tests](https://github.com/eisop/jspecify-conformance), so we directly implemented a `DetailedTestDiagnostic` class in EISOP that can also be used here.
Not strictly necessary to do, but would reduce the code in this repository and protect against changes in the error format.


wmdietl:
Sorry, I just noticed that in https://github.com/eisop/checker-framework/pull/758 I did break how this project parses the detailed diagnostics itself... I've filed https://github.com/eisop/checker-framework/issues/809 and can look into fixing this.
It would seem simpler to just merge this PR though.

cpovirk (approved):

src/test/java/tests/NullSpecTest.java
```diff
@@ -117,10 +117,8 @@ public TypecheckResult adjustTypecheckResult(TypecheckResult testResult) {
 
     for (ListIterator<TestDiagnostic> i = unexpected.listIterator(); i.hasNext(); ) {
       TestDiagnostic diagnostic = i.next();
-      DetailMessage detailMessage = DetailMessage.parse(diagnostic, null);
-      if (detailMessage.hasDetails()) {
-        // Replace diagnostics that can be parsed with DetailMessage diagnostics.
-        i.set(detailMessage);
+      if (diagnostic instanceof DetailedTestDiagnostic) {
```
wmdietl, line 120: As the normal tests are executed without `-Adetailedmsgtext`, this is likely a dead branch. I'll address this in a separate PR.

PR #187, opened 2024-07-12
Update to gradle 8.9

wmdietl:


cpovirk (approved):
It might not matter, since we're hopefully nearing the point at which we can merge `main-eisop` into `main`, but you might consider merging this PR onto `main` instead of `main-eisop`. But if that will mainly just complicate your life, then don't worry about it :)

PR #194, opened 2024-07-26
Report all errors as [nullness].

cpovirk:


wmdietl (commented):
This depends on https://github.com/eisop/checker-framework/pull/816, which isn't in a release yet.
So you get `NullSpecChecker.java:72: error: method does not override or implement a method from a supertype`.
You could change CI to again build a local clone of EISOP and use that.

wmdietl (approved):
#196 includes a local CF in CI so this compiles now.

PR #196, opened 2024-07-26
Build a local `eisop/checker-framework` for CI

wmdietl:


cpovirk (approved):
Thanks!

build.gradle
```diff
@@ -98,7 +98,7 @@ dependencies {
 // If built with `--include-build path/to/checker-framework` then
 // assemble checker-framework when assembling the reference checker.
 if (checkerFramework != null) {
-    assemble.dependsOn(checkerFramework.task(":assembleForJavac"))
+    assemble.dependsOn(checkerFramework.task(":checker:assembleForJavac"))
```
wmdietl: I'm a bit confused why this is needed with the `include-build` and wasn't before... it manually works to run `./gradlew assembleForJavac` on the top-level of a `checker-framework` checkout.

cpovirk: I don't know, but thank you for figuring it out.

Issue #193, opened 2024-07-23
New errors (maybe related to `? super`) from eisop/checker-framework#746


cpovirk:
Using [the command given here](https://github.com/jspecify/jspecify-reference-checker/pull/178#issuecomment-2243407154) to test with `main-eisop`:

```
/tmp/tmp.OBavpODHqE/guava/guava/src/com/google/common/collect/Multisets.java:1123: error: [argument.type.incompatible] incompatible argument for parameter o of Collections.nCopies.
        entry -> Collections.nCopies(entry.getCount(), entry.getElement()).spliterator(),
                                                                       ^
  found   : E*
  required: E
```

`E*` should be a valid substitute for `E` in lenient mode.

```
/tmp/tmp.OBavpODHqE/guava/guava/src/com/google/common/collect/TreeMultimap.java:224: error: [assignment.type.incompatible] incompatible types in assignment.
    keyComparator = requireNonNull((Comparator<? super K>) stream.readObject());
                                  ^
  found   : Comparator<capture#767 of ? super K>
  required: Comparator<? super K>
```

I'm not totally sure what's up with that one, nor even that these two errors are closely related.

Previous discussions (which also touched on the problem that `SequencedCollection` and friends are unannotated):
- https://github.com/jspecify/jspecify-reference-checker/pull/178#issuecomment-2072654491
- https://github.com/eisop/checker-framework/pull/746#issuecomment-2239583272
- https://github.com/jspecify/jspecify-reference-checker/issues/192

I have promised to try to produce a smaller repro of this problem. For now, I'm filing this issue as a placeholder.

wmdietl:
I tried minimizing a test case with `? super TV` that doesn't depend on the JDK:

````
> cat Issue193.java 
import org.jspecify.annotations.NullMarked;
import org.jspecify.annotations.Nullable;

class Box<B extends @Nullable Object> {}

interface MyList<T extends @Nullable Object> {
  boolean removeIf(Box<? super T> b);
}

@NullMarked
interface ListSubclass<Z extends @Nullable Object> extends MyList<Z> {
  boolean removeIf(Box<? super Z> b);
}

> ./demo Issue193.java
Issue193.java:12: error: [override.param.invalid] Incompatible parameter type for b.
  boolean removeIf(Box<? super Z> b);
                                  ^
  found   : Box<? super Z>
  required: Box<? super Z*>*
  Consequence: method in ListSubclass<Z>
[...]

PR #197, opened 2024-08-08
Fix capture conversion in lenient mode.

cpovirk:
...by always performing capture conversion in _strict_ mode, where there
is a better-defined hierarchy (non-null < unspecified < nullable, with
no "but also unspecified < non-null if you want).

Fixes https://github.com/jspecify/jspecify-reference-checker/issues/193

### DIFF
--- a/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
+++ b/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
@@ -532,6 +532,14 @@ final class NullSpecAnnotatedTypeFactory
   }
 
   @Override
+  public AnnotatedTypeMirror applyCaptureConversion(
+      AnnotatedTypeMirror type, TypeMirror typeMirror) {
+    return this == withLeastConvenientWorld
+        ? super.applyCaptureConversion(type, typeMirror)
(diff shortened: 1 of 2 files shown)


wmdietl (approved):
Thanks for the detailed discussion of the fix in #193!
The fix makes sense. I'm wondering whether to add the test from https://github.com/jspecify/jspecify-reference-checker/issues/193#issuecomment-2274266977 either as a regression test here or as a conformance test.

cpovirk:
Good point. I'm adding a regression test for now to match what we've done before, but the `regression` directory should obviously be one source of possible conformance tests in the future. (There will be an interesting discussion to be had about which kinds of bugs justify per-checker tests vs. those which justify conformance tests. Maybe the right outcome will be that _everything_ becomes a conformance test but that the borderline cases get thrown into a `junk-drawer` directory where no one expects organization? Maybe that wouldn't even be part of "the conformance tests" per se, just a part of the same repo for convenient integration? Or maybe that's all a bad idea.)

I've confirmed that the new test passes with `JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64 ./gradlew test --include-build ../jspecify --include-build ../checker-framework`. (And I "intentionally" verified that it can fail by checking yet again whether I could remove the wonkyness workaround :) Yet again, I found that I can't.)

PR #201, opened 2024-08-22
Use the new `NoContractsFromMethod`

wmdietl:
Depends on https://github.com/eisop/checker-framework/pull/878
Also tests whether the EISOP CI picks up this corresponding branch in the reference checker.

### DIFF
--- a/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
+++ b/src/main/java/com/google/jspecify/nullness/NullSpecAnnotatedTypeFactory.java
@@ -104,12 +104,10 @@ import org.checkerframework.framework.type.treeannotator.TreeAnnotator;
 import org.checkerframework.framework.type.typeannotator.TypeAnnotator;
 import org.checkerframework.framework.type.visitor.AnnotatedTypeVisitor;
 import org.checkerframework.framework.util.AnnotationFormatter;
-import org.checkerframework.framework.util.Contract.ConditionalPostcondition;
-import org.checkerframework.framework.util.Contract.Postcondition;
-import org.checkerframework.framework.util.Contract.Precondition;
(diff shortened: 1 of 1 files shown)


wmdietl:
EISOP CI build correctly picks up this branch and is successful.
I'll merge the two PRs together once they're approved.

In an ideal world this PR would pick up the corresponding EISOP branch...

cpovirk (approved):
Nice.

wmdietl:
In CI I notice this failure:

````
FAILURE: Build failed with an exception.

* What went wrong:
Circular dependency between the following tasks:
:jspecify:classes
\--- :jspecify:compileJava
     \--- :jspecify:jar
          +--- :jspecify:classes (*)
          +--- :jspecify:compileJava (*)
          +--- :jspecify:compileJava9Java
          |    \--- :jspecify:jar (*)
          \--- :jspecify:java9Classes
               \--- :jspecify:compileJava9Java (*)

(*) - details omitted (listed previously)
````

The same problem can be reproduced with `main`, meaning it doesn't depend on this PR and doesn't depend on changes in EISOP.

I can reproduce it with jspecify-reference-checker in the `main` branch using an `--include-build ../jspecify` configuration.
However, I didn't figure out how to trigger the circular dependency directly in the `jspecify` build.

When I go into `jspecify` and execute `git checkout 0394cfed29ae86d75ad327dc9df9697e0b9b90a6`, that is, if I use the state before https://github.com/jspecify/jspecify/pull/605, the build of the reference checker succeeds.

So something about https://github.com/jspecify/jspecify/pull/605 is breaking the build of the reference checker.

@cpovirk Any idea?

wmdietl:
I'll merge this PR to make `main-eisop` compile against the updated EISOP. I'll turn the previous comment into a new issue.

Issue #202, opened 2024-08-23
Circular dependency with `:jspecify:classes`


wmdietl:
In CI of #201 I notice this failure:

````
FAILURE: Build failed with an exception.

* What went wrong:
Circular dependency between the following tasks:
:jspecify:classes
\--- :jspecify:compileJava
     \--- :jspecify:jar
          +--- :jspecify:classes (*)
          +--- :jspecify:compileJava (*)
          +--- :jspecify:compileJava9Java
          |    \--- :jspecify:jar (*)
          \--- :jspecify:java9Classes
               \--- :jspecify:compileJava9Java (*)

(*) - details omitted (listed previously)
````

The same problem can be reproduced with `main`, meaning it doesn't depend on PR #201 and doesn't depend on changes in EISOP.

I can reproduce it with jspecify-reference-checker in the `main` branch using an `--include-build ../jspecify` configuration.
However, I didn't figure out how to trigger the circular dependency directly in the `jspecify` build.

When I go into `jspecify` and execute `git checkout 0394cfed29ae86d75ad327dc9df9697e0b9b90a6`, that is, if I use the state before https://github.com/jspecify/jspecify/pull/605, the build of the reference checker succeeds.

So something about https://github.com/jspecify/jspecify/pull/605 is breaking the build of the reference checker.

@cpovirk Any idea?

_Originally posted by @wmdietl in https://github.com/jspecify/jspecify-reference-checker/issues/201#issuecomment-2307922153_

cpovirk:
Thanks. I should probably just revert that PR.

PR #204, opened 2024-08-23
Update to gradle 8.10

wmdietl:
This updates the `main` branch and should illustrate that #202 is not just an `main-eisop` issue.


cpovirk (approved):

netdpb:
Looks like a build cycle?

wmdietl:
#202 was fixed by https://github.com/jspecify/jspecify/pull/609.

PR #215, opened 2024-12-22
Update to Gradle 8.12

wmdietl:


build.gradle
```diff
@@ -109,6 +109,12 @@ if (jspecify != null) {
     assemble.dependsOn(jspecify.task(':assemble'))
 }
 
+// Enable exec/javaexec
```
wmdietl, line 112: I'm happy for suggestions for how to do this in a nicer way... The [upgrade instructions](https://docs.gradle.org/8.12/userguide/upgrading_version_8.html#deprecated_project_exec) didn't really help me a lot.

cpovirk: Sadly but unsurprisingly, I'm clueless. Now that I know to look for `ExecOperations`, I see https://github.com/gradle/gradle/issues/30822, which seems encouraging, I think?

cpovirk (approved):
Thanks for taking care of this.

build.gradle
wmdietl: Thanks, here I was doing the right (oddly verbose) thing but in `settings.gradle` I wasn't actually executing the task. Now it both compiles and seems to do the right thing.

PR #217, opened 2025-01-06
Add subdirectories to dependabot

wmdietl:


wmdietl:
I looked a bit into dependabot configuration and `target-branch` would not help us, as only the default branch is checked by dependabot. So this change will only be useful once `main-eisop` becomes `main`.

On the `main` branch, [build.gradle](https://github.com/jspecify/jspecify-reference-checker/blob/a9958c5cc11906b16027e5d105501216d52b3bcc/build.gradle#L11) doesn't seem to get updated, neither the versions in [settings.gradle](https://github.com/jspecify/jspecify-reference-checker/blob/a9958c5cc11906b16027e5d105501216d52b3bcc/settings.gradle#L25).

I don't know if we're missing some general dependabot configuration issue...

cpovirk (approved):

